PR #32, opened 2016-12-30
DBZ-3 Adds docs for the PostgreSQL connector

hchiorean:


docs/connectors/postgresql.asciidoc
```diff
@@ -2,5 +2,1044 @@
...
+----
+
+If the `database.server.name` configuration property has the value `PostgreSQL-server-1`, every change event for the `customers` table while it has this definition will feature the same key structure, which in JSON looks like this:
+
+[source,json,indent=0]
+----
+  {
+    "schema": {
+      "type": "struct",
+      "name": "PostgreSQL-server-1.public.customers.Key"
```
rhauch: This doesn't comply with Avro Schema names, does it? Won't the `-` characters be replaced with `_`? If so, wouldn't it be easier/simpler to have the example where everything matches?

docs/connectors/postgresql.asciidoc
```diff
@@ -2,5 +2,1044 @@
...
 :linkattrs:
 :icons: font
+:source-highlighter: highlight.js
+:debezium-version: 0.4.0
 
-This connector is https://issues.jboss.org/browse/DBZ-3[incomplete]. Want to help us work on it? link:/docs/contribute[Learn how].
+Debezium's PostgreSQL Connector can monitor and record the row-level changes in the schemas of a PostgreSQL database. 
+
+The first time it connects to a PostgreSQL server/cluster, it reads a consistent snapshot of all of the schemas. When that snapshot is complete, the connector continuously streams the changes that were committed to PostgreSQL 9.6 or later and generates corresponding insert, update and delete events. All of the events for each table are recorded in a separate Kafka topic, where they can be easily consumed by applications and services.
+
```
hchiorean: All the tests (including our Docker image) were done on 9.6

`9.4` was the first version where logical decoding was introduced, but it had several bugs which were fixed subsequently in 9.5 and 9.6

So, in theory we can say `9.4`, but IMO it's safer to start from `9.6`

docs/connectors/postgresql.asciidoc
```diff
@@ -2,5 +2,1044 @@
...
+The third snapshot mode ensures the connector *never* performs snapshots. When a new connector is configured this way, if will either continue streaming changes from a previous stored offset or it will start from the point in time when the PostgreSQL logical replication slot was first created on the server.
+
+The final snapshot mode - *initial only* - will simply perform a database snapshot and stop streaming any other changes afterwards. If a previous snapshot was in effect but had not completed successfully, the connector will restart the snapshot process each time until this completes successfully.
+
```
hchiorean: there is no easy way to tell is an `LSN` number exists or not on the server.

The way logical replication streaming works is you give the server an `LSN` and it will simply either use it (if it still has the segments) or will silently discard it and start streaming changes from the latest available `LSN`. There is no "feedback" in the logical replication streaming protocol to inform the client that the `LSN` is no longer available.

This could be implemented (probably requires looking in some system tables), but it also means changes to the connector and some additional database queries and logic. So if the `when_needed` mode is required, I'd rather have that as a feature request.

docs/connectors/postgresql.asciidoc
hchiorean: fair point; this was copied from the MySQL connector docs :)

hchiorean:
amended the commit fixing the `server-name` to contain valid Avro schema chars

hchiorean:
added more information about clustering and potential failures

rhauch:
Thanks for the documentation, @hchiorean. I've added your commit to my pull request #34 that will contain all of the website changes for the upcoming 0.4.0 release.

PR #46, opened 2017-06-01
DBZ-222 Updated the docs for POINT

omarsmak:
Added the documentation for POINT data type per [DBZ-222](https://issues.jboss.org/browse/DBZ-222). 

P/S: Sorry for the formatting issues.

FYI @gunnarmorling

### DIFF
--- a/docs/connectors/mysql.asciidoc
+++ b/docs/connectors/mysql.asciidoc
@@ -993,7 +993,7 @@ However, when `decimal.handling.mode` configuration property is set to `double`,
 [[spatial-types]]
 ==== Spatial Data Types
 
-The MySQL connector also has limited support for some of the following https://dev.mysql.com/doc/refman/5.7/en/spatial-datatypes.html[spatial data types]:
+As of version 0.5.1, the MySQL connector also has limited support for some of the following https://dev.mysql.com/doc/refman/5.7/en/spatial-datatypes.html[spatial data types]:
 
 [cols="20%a,15%a,30%a,35%a",width=150,options="header,footer",role="table table-bordered table-striped"]
 |=======================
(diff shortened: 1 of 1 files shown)


rhauch:
It'd really be great if the PR didn't have all of the formatting changes since they hide the real meaningful changes.

gunnarmorling:
Hey, I generally think the removal of trailing whitespace is a good thing, but I found it a good practice to have this kind of change in a separate commit, which reviewers then can easily remove.

But GitHub actually has a little trick to help with that, append "w=1" to a diff URL and it'll ignore any whitespace change: https://github.com/debezium/debezium.github.io/pull/46/files?w=1

rhauch:
@gunnarmorling Hey, neat trick! I do agree that it's better to get rid of the trailing whitespace, as long as its in a separate commit. But that `w=1` trick will certainly make things easier!

docs/connectors/mysql.asciidoc
```diff
@@ -978,18 +978,36 @@ However, when `decimal.handling.mode` configuration property is set to `double`,
...
 |`FLOAT64`
 |
 |
 
 |=======================
 
+[[spatial-types]]
+==== Spatial Data Types
+
+The MySQL connector also has limited support for some of the following https://dev.mysql.com/doc/refman/5.7/en/spatial-datatypes.html[spatial data types]:
```
rhauch: Actually, maybe it be better to say "As of version 0.5.1, the MySQL connector ..."

rhauch:
@omarsmak I do have one suggestion about mentioning the Debezium version where the geometry types are supported. WDYT?

docs/connectors/mysql.asciidoc
omarsmak: I think it'd sound better, will update the PR then

gunnarmorling:
Rebased and applied; thanks, @omarsmak!

PR #49, opened 2017-07-11
Add note for non-Linux users about IP binding

jpechane:


gunnarmorling:
Thanks!

Btw. for new contents to be written, I'd suggest to avoid very long lines.

They make it harder to view changes in the diff view (e.g. no word-diff highlighting on GitHub). Instead of a fixed length, though (which forces constant re-wrapping after changes), we've made good experiences with using the rule "one sentence or line of thought per line" in the Hibernate docs. I.e. shorter sentences will be on their own line, for longer ones one would move to a new line after a comma.

E.g. like so:

    Btw. for new contents to be written, I'd suggest to avoid very long lines.
    
    They make it harder to view changes in the diff view (e.g. no word-diff highlighting on GitHub).
    Instead of a fixed length, though (which forces constant re-wrapping after changes),
    we've made good experiences with using the rule "one sentence or line of thought per line" in the Hibernate docs.     
    I.e. shorter sentences will be on their own line, for longer ones one would move to a new line after a comma.

PR #54, opened 2017-07-28
Avro converter documentation

jpechane:


docs/configuration/avro.asciidoc
```diff
@@ -0,0 +1,65 @@
+= Avro Serialization
+:awestruct-layout: doc
+:linkattrs:
+:icons: font
+:source-highlighter: highlight.js
+
+Debezium by default serializes data events into a free-form JSON format. It is possible to use https://avro.apache.org/[Avro] serialization that will use binary encoding for the data and associate a schema identifier with the message to allow safe schema evolution over a longer time period.
+
+Debezium plug-ins will typically use Avro serializations in two different scenarios
```
rhauch: We need to be careful about how we describe Debezium vs Kafka Connect, and this paragraph makes it more confusing for people to understand the distinction. I suggest replacing the first 2 paragraphs with something like the following:

> Debezium connectors are used with the Kafka Connect framework to capture changes in databases and generate change events. The Kafka Connect workers then apply to each of the messages generated by the connector the transformations configured for the connector, serialize each message key and value into a binary form using the worker's http://docs.confluent.io/3.2.1/connect/concepts.html#connect-converters[_converters_], and finally write each messages into the correct Kafka topic. 
>
> The converters are specified in the the Kafka Connect worker configuration, and the same converters are used for all connectors deployed to that worker's cluster. Kafka Connect comes with a _JSON converter_ that serialize the message keys and values into JSON documents, and the JSON converter can be configured to include or exclude the message schema using the (`key.converter.schemas.enable` and `value.converter.schemas.enable`) properties. Our link:/docs/tutorial[tutorial] shows what the messages look like when both payload and schemas are included, but the schemas make the messages very verbose. If you want your messages serialized with JSON, consider setting these properties to `false` to exclude the verbose schema information.
>
> Another option is to serialize the message keys and values using https://avro.apache.org/[Apache Avro]. The Avro binary format is extremely compact and efficient, and Avro schemas make it possible to ensure that the messages have the correct structure. Avro's schema evolution mechanism makes it possible to evolve the schemas over time, which is essential for Debezium connectors that dynamically generate the message schemas to match the structure of the database tables. Over time, the change events captured by Debezium connectors and written by Kafka Connect into a topic may have different versions of the same schema, and Avro serialization makes it far easier for consumers to adapt to the changing schema.
>
> Confluent provides http://docs.confluent.io/3.2.1/schema-registry/docs/index.html[several components] that work with Avro:
> * An Avro Converter that can be used in Kafka Connect workers to map the Kafka Connect schemas into Avro schemas and to then use those Avro schemas to serialize the message keys and values into the very compact Avro binary form.
> * A Schema Registry that tracks all of the Avro schemas used in Kafka topics, and where the Avro Converter sends the generated Avro schemas. Since the Avro schemas are stored in this registry, each message need only include a tiny _schema identifier_. This makes each message even smaller, and for an I/O bound system like Kafka this means more total throughput of the producers and consumers.
> * Avro _serdes_ (serializers and deserializers) for Kafka producers and consumers. Any Kafka consumer applications you write to consume change events can use the Avro serdes to deserialize the changes events.
>
> These Confluent components are open source, and you can install them into any Kafka distribution and use them with Kafka Connect. However, Confluent also provides a https://www.confluent.io/product/confluent-open-source/[Confluent Open Source Platform] that includes the standard Kafka distribution as well as these and other Confluent open source components, including several source and sink connectors. Some Docker images for Kafka Connect also contain the Avro converter. This includes recent http://debezium.io/docs/docker/[Debezium Docker images] that include the Debezium connectors, Kafka Connect, and the Avro converter.

docs/configuration/avro.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+* Deploy a https://github.com/confluentinc/schema-registry[Schema Registry] instance
+* Use these properties to configure Apache Connect instance
+[source]
+----
+key.converter=io.confluent.connect.avro.AvroConverter
+key.converter.schema.registry.url=http://localhost:8081
+value.converter=io.confluent.connect.avro.AvroConverter
+value.converter.schema.registry.url=http://localhost:8081
+----
+Note: In addition to setting key/value converters it is necessary to set internal key/value converters due to a https://issues.apache.org/jira/browse/KAFKA-3988[known issue].
```
rhauch: Actually, we recommend to _always_ use JSON converters for the internal keys and values. The volume of messages is small, and the readability helps if you ever need to see/edit the stored configurations and offsets.

docs/configuration/index.html.haml
```diff
@@ -0,0 +1,13 @@
+---
+layout: doc
+title: Configuration
+title_link: /docs/configuration
+---
+
+%p
+  This section provides information about a cross-cutting configuration options common for all plug-ins like logging, serialization et al.
```
rhauch: "... about configuring the Kafka Connect workers used with Debezium connectors.

gunnarmorling:
@jpechane where are we with this one? Seems you still need to incorporate the remarks of @rhauch?

jpechane:
I'll finish it by tomorrow, no objections to @rhauch comments

gunnarmorling:
Ok, thanks. There are some typos in the amendment, so either you fix those yourself or push another commit with the change to the PR and I can take it from there.

gunnarmorling:
@rhauch interesting note about the ability to turn off the addition of schemas to JSON messages. Do you know whether this works in conjunction with the Schema Registry? Or is that one only supported when using Avro-style messages?

rhauch:
The JSON converter configuration has always been a part of the Apache Kafka and is independent of the Avro converter.

gunnarmorling:
I understand that the JSON and Avro converters are independent. But my question (or, rather, idea) was that the JSON format still could be used the schema registry. I.e. just as with Avro, instead of having the schema inline within the message, there could be a reference in JSON documents to the schema id in the registry.

In other words, the fact a) that the schema is stored inline or externally and b) the serialized format (JSON vs. Avro) are two separate things and it seems generally possible to support any combination of inline/external schema storage and serialization format. It seems not supported right now, but perhaps an interesting feature request for Connect and the schema registry?

gunnarmorling:
Rebased and applied. Thanks @jpechane and @rhauch!

PR #55, opened 2017-08-16
DBZ-325 Add docs and menu entries for new topic routing

xenji:
The changes cover the docs of https://github.com/debezium/debezium/pull/267. As my origins are non-english, please let a native speaker review the grammar and wording of it ... I'm looking at you, @rhauch ;)

### DIFF
--- a/docs/configuration/topic-routing.asciidoc
+++ b/docs/configuration/topic-routing.asciidoc
@@ -30,7 +30,8 @@ that make up the primary key, the events for that row in shard 1 would have the 
 even though globally they are different rows. So, in addition to changing the topic name, we may also want to _modify 
 the event key_ to add a field that makes the key globally unique.
 
-This SMT lets you specify how you want to choose the new topic name and then specify how to modify the change event key to ensure it is globally unique.
+This SMT lets you specify how you want to choose the new topic name and then specify how to modify the change event key 
+to ensure it is globally unique.
 
 == Topic Names
 
@@ -46,7 +47,7 @@ transforms.Reroute.topic.replacement=$1
 ---
 
 The configuration above will match a topic named `db1.foo__bar` and replace it with `db1.foo`, but keep the key's name
-`db1.foo__bar.Key`. If you intend to change also the key's name, please follow the instructions in (key-fields)[Key Fields]
+`db1.foo__bar.Key`. If you intend to change also the key's name, please follow the instructions in (key-fields)[Key Fields].
 
 == Key Fields
 
--- a/docs/configuration/topic-routing.asciidoc
+++ b/docs/configuration/topic-routing.asciidoc
@@ -1,0 +1,55 @@
+= Topic Routing
+:awestruct-layout: doc
+:linkattrs:
+:icons: font
+:source-highlighter: highlight.js
+
+Debezium enables you to re-reoute the emitted change before the message reaches the converter using a 
+single message transformation, or https://kafka.apache.org/documentation/#connect_transforms[SMT]. 
+The SMT provided by Debezium enables you to rewrite the topic and the key according to a regular expression 
+and a replacement pattern, configurable per instance of Debezium.
+
+The implementation does not care about the sanity of the change, this is in the responsibility of the user.
+
+== Use-Cases
+
+=== Virtual Tables
+
+A logical table consists of one or more physical tables with the same table structure. A common use case is sharding. The
+two physical tables `db_shard1.my_table` and `db_shard2.my_table` together form one logical table.
+
+== Topic Names
+
+Below is an example for a configuration which replaces a part of the table in the topic with another string, allowing
+two tables to emit changes to the same topic, keeping their different key namings.
+
+[source]
+---
+transforms=Reroute
+transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
+transforms.Reroute.topic.regex=(.*)__bar$
+transforms.Reroute.topic.replacement=$1
+---
+
+The configuration above will match a topic named `db1.foo__bar` and replace it with `db1.foo`, but keep the key's name
+`db1.foo__bar.Key`. If you intend to change also the key's name, please follow the instructions in (key-fields)[Key Fields]
+
+== Key Fields
+
+Using only the topic replacement does not touch the key's field. In order to make it completly transparent to the
+consumer, whether the event came from one or another table, a key field change is also required. 
+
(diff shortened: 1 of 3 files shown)


docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+Debezium enables you to re-reoute the emitted change before the message reaches the converter using a 
+single message transformation, or https://kafka.apache.org/documentation/#connect_transforms[SMT]. 
+The SMT provided by Debezium enables you to rewrite the topic and the key according to a regular expression 
+and a replacement pattern, configurable per instance of Debezium.
+
+The implementation does not care about the sanity of the change, this is in the responsibility of the user.
+
+== Use-Cases
+
+=== Virtual Tables
```
gunnarmorling: Let's use one term, "logical table" or "virtual table".

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+[source]
+---
+transforms=Reroute
+transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
+transforms.Reroute.topic.regex=(.*)__bar$
+transforms.Reroute.topic.replacement=$1
+---
+
+The configuration above will match a topic named `db1.foo__bar` and replace it with `db1.foo`, but keep the key's name
+`db1.foo__bar.Key`. If you intend to change also the key's name, please follow the instructions in (key-fields)[Key Fields]
```
gunnarmorling: The second sentence is missing a dot.

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+== Key Fields
+
+Using only the topic replacement does not touch the key's field. In order to make it completly transparent to the
```
gunnarmorling: s/completly/completely/

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
+transforms.Reroute.topic.regex=(.*)__bar$
+transforms.Reroute.topic.replacement=$1
+transforms.Reroute.key.field.regex=(.*)__bar$
+transforms.Reroute.key.field.replacement=$1
+transforms.Reroute.key.field.name=__dbz__
+---
+
+The configuration above applies the same replacement to the key field that has already been applied to the topic name.
+In addtion, the `key.field.name` configuration can be used to declare a field which will contain the original key name
```
gunnarmorling: s/addtion/addition/

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+for later use. The configuration shows the default value of `$$__dbz__$$`, which can be changed to a user provided value.
```
gunnarmorling: In the example above you don't have the leading/trailing $ signs. Also I'd suggest to use a more descriptive name to make the example for comprehensible. How about mentioning the default value, too, which will be used if no value is explictly given (__dbz__physicalTableIdentifier)?

rhauch:
> Thinking about this feature in general, I'm wondering, shouldn't we also have a way to keep the original table name somehow? E.g. for a sharding use case, I'd expect that each shard table has the same key name, e.g. "id", and then you may want to add the table so to know the originating shard later on. Or am I missing something?

The SMT already adds to the key fields for the names of the database and table where the event originated. This is necessary to ensure that the event key remains unique.

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+and a replacement pattern, configurable per instance of Debezium.
+
+The implementation does not care about the sanity of the change, this is in the responsibility of the user.
+
+== Use-Cases
+
+=== Virtual Tables
+
+A logical table consists of one or more physical tables with the same table structure. A common use case is sharding. The
+two physical tables `db_shard1.my_table` and `db_shard2.my_table` together form one logical table.
```
rhauch: Maybe "A common use case is sharding, where for example two physical tables..."?

docs/configuration/topic-routing.asciidoc
rhauch: Actually, I'd suggest:

> This example configuration defines `source_table` as the name of the new field we'll add to the change event's key structure. We'd like to have the value of this new field to contain the name of the source table where the event originated, so we use the `topic.regex` property to define a regular expression that captures the source table name from the original topic, and the `key.field.replacement` property specifies how the new field's value should be computed from the captured groups, which in our example is the first capture group that contains the source table name.
>
> Obviously you will want to choose the name of the new key field carefully so that it doesn't clash with an existing field in the primary key.

docs/configuration/topic-routing.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+[source]
+---
+transforms=Reroute
+transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
+transforms.Reroute.topic.regex=(.*)__bar$
+transforms.Reroute.topic.replacement=$1
+transforms.Reroute.key.field.regex=(.*)__bar$
+transforms.Reroute.key.field.replacement=$1
+transforms.Reroute.key.field.name=__dbz__
```
rhauch: See below, but I suggest changing the property value to `source_table`.

docs/configuration/topic-routing.asciidoc
xenji: The leading and trailing `$$` are for asciidoc to ignore the double-underscore as a formatting hint.

gunnarmorling: +1 for Randall's suggestion above. I'd only add one more sentence before "We'd like to have...":

> If no name for the additional key field is specified, it will be named `__dbz__physicalTableIdentifier`.

gunnarmorling:
Applied, thanks a lot, @xenji. I've added two more commits:

* Further clarification around `key.field.regex` and `replacement`; it is applied to the original topic name and the replacement is used as the field value. It is not related to the original key name. Also the field is actually always added to the transformed messages (using the default if no name is explicitly given).
* Using the "one sentence/thought per line" wrapping style. This will facilitate future changes and diffing them, without having to re-wrap large parts of the document after insertions for example.

PR #56, opened 2017-08-16
DBZ-330 add hint for property to jdbc url passthrough

xenji:


docs/connectors/mysql.asciidoc
```diff
@@ -1381,6 +1381,8 @@ The connector also supports _pass-through_ configuration properties that are use
 
 For example, the following connector configuration properties can be used to http://kafka.apache.org/documentation.html#security_configclients[secure connections to the Kafka broker]:
 
+In addition to the _pass-through_ to the Kafka producer and consumer, the properties starting with `database.`, e.g. `database.tinyInt1isBit=false` are passed to the JDBC URL.
```
gunnarmorling: In hindsight it seems we should have used distinct prefixes for the different property spaces. Now we pass the Kafka props to the JDBC driver, too...

rhauch: Logged as https://issues.jboss.org/browse/DBZ-333

PR #66, opened 2017-09-19
DBZ-347 Configurable behaviour for unknown DDL

jpechane:


docs/connectors/mysql.asciidoc
```diff
@@ -1365,6 +1365,9 @@ The following _advanced_ configuration properties have good defaults that will w
 |`4`
 |The maximum number of times that the connector should attempt to read persisted history data before the connector recovery fails with an error. The maximum amount of time to wait after receiving no data is `recovery.attempts` x `recovery.poll.interval.ms`.
 
+|`database.history.skip.unparseable.ddl`
+|`false`
+|Boolean value that specifies if connector should ignore malformed or unknown database statements or stop processing and let operator to fix the issue. The safe default is `true`. Skipping should be used only with care as it can lead to data loss or mangling when binlog is processed.
```
gunnarmorling: LGTM. Only in terms of formatting, my preference still would be to go for the "one sentence per line" style. Long lines as this make diffs much harder to digest down the road.

PR #68, opened 2017-09-19
DBZ-306 update docs to reflect changes in event key’s payload

hpgrahsl:
also add several examples showing how different _id field types will get encoded as JSON strings

### DIFF
--- a/docs/connectors/mongodb.asciidoc
+++ b/docs/connectors/mongodb.asciidoc
@@ -200,15 +200,16 @@ The `schema` portion of the key contains a Kafka Connect schema describing what 
 This example used a document with an integer identifier, but any valid MongoDB document identifier (including documents) will work. The value of the `id` field in the payload will simply be a string representing a MongoDB extended JSON serialization (strict mode) of the original document's `_id` field. Find below a few examples showing how `_id` fields of 
 different types will get encoded as the event key's payload:
 
-* `_id` of type integer e.g. 1234 -> `{ "id" : "1234" }`
-* `_id` of type float e.g. of 12.34 -> `{ "id" : "12.34" }`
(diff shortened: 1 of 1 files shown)


docs/connectors/mongodb.asciidoc
```diff
@@ -190,14 +190,25 @@ Every change event for the `customers` collection will feature the same key stru
...
 
-This example used a document with an integer identifier, but any valid MongoDB document identifier (including documents) will work. The value of the `_id` field in the payload will simply be the string representation of that value.
+This example used a document with an integer identifier, but any valid MongoDB document identifier (including documents) will work. The value of the `id` field in the payload will simply be a string representing a MongoDB extended JSON serialization (strict mode) of the original document's `_id` field. Find below a few examples showing how `_id` fields of 
+different types will get encoded as the event key's payload:
+
+* `_id` of type integer e.g. 1234 -> `{ "id" : "1234" }`
+* `_id` of type float e.g. of 12.34 -> `{ "id" : "12.34" }`
+* `_id` of type string e.g. "1234" -> `{ "id" : "\"1234\"" }`
+* `_id` of type document e.g. { "hi" : "kafka", "nums" : [10.0, 100.0, 1000.0] } -> `{ "id" : "{\"hi\" : \"kafka\", \"nums\" : [10.0, 100.0, 1000.0]}" }`
```
rhauch: Would it be better to use a table with columns "Type", "MongoDB `_id` Value", "Key's payload"? If so, the last two lines below should be moved into the table as well.

hpgrahsl: Fine for me if you prefer tabular layout. Here we go

gunnarmorling:
Merged; thanks, @hpgrahsl!

PR #72, opened 2017-09-22
DBZ-371 Introducing global variables for latest Debezium version and …

gunnarmorling:
…Docker label; referring to these variables in many places

https://issues.jboss.org/browse/DBZ-371

### DIFF
--- a/docs/install.asciidoc
+++ b/docs/install.asciidoc
@@ -2,7 +2,6 @@
 :awestruct-layout: doc
 :linkattrs:
 :icons: font
-:confluent-platform-version: 3.2.0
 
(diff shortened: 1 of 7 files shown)


docs/install.asciidoc
```diff
@@ -2,7 +2,6 @@
 :awestruct-layout: doc
 :linkattrs:
 :icons: font
-:debezium-version: 0.5.0
 :confluent-platform-version: 3.2.0
```
jpechane: This one should be extracted too and promoted to 3.3.0

gunnarmorling: Done.

jpechane:
Very useful, thanks!

PR #74, opened 2017-09-25
Blogspot for DB->DB replication

jpechane:


blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+* Apache ZooKeeper
+* Apache Kafka
+* An enriched Kafka Connect image with changes
+** PostgreSQL JDBC driver placed into `/kafka/libs` directory
+** https://docs.confluent.io/current/connect/connect-jdbc/docs/index.html[Kafka Connect JDBC Connector] placed into `/kafka/connect/kafka-connect-jdbc` directory
+* Pre-populated MySQL used in our link:docs/tutorial[tutorial]
+* Empty PostgreSQL
+
+There are certain impedances between Debezium MySQL Connector and Kafka Connect JDBC connector
```
rhauch: Did you mean to use the term "impedance mismatch"? I suggest perhaps explaining things a bit more:

> The Debezium MySQL Connector was designed to specifically capture database changes and provide as much information as possible about those events beyond just the new state of each row. Meanwhile, the Confluent JDBC Sink Connector was designed to simply convert each message into a database insert/upsert based upon the structure of the message. So, the two connectors have different structures for the messages, but they also use different topic naming conventions and behavior of representing deleted records.
> These mismatches in structure and behavior will be common when using connectors that were not designed to work together. But this is something that we can easily deal with, and we discuss how in the next few sections.

IMO, the first sentence of the last paragraph is really important, and should be stated explicitly.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+* event format
+* delete records/tombstone records handling
+* topic naming
+
+We are going to go through all those issues in detail.
+
+=== Event format
+Debezium emits events in a complex format that contains two record values: a value of the row record before the change and value of the row record after the change is applied.
+Further metadata are also included and the result is so called `Envelope` format.
```
rhauch: I'd suggest:

> Debezium emits events in a complex format that contains all of the information about the captured data change: the type of operation, source metadata, the timestamp the event was processed by the connector, and state of the the row before and after the change was made. Debezium calls this structure an "envelope":

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+		"field2" : "oldvalue2"
+	},
+	"after" : {
+		"field1" : "newvalue1",
+		"field2" : "newvalue2"
+	}
+}
+----
+
+Kafka Connect JDBC connector (as most of the others) emits outgoing and expects incoming events in a simple format that is generally a new row record.
```
rhauch: > Many other Kafka Connect source connectors don't have the luxury of knowing this much about the changes, and instead use a simpler model where each message directly represents the _after_ state of the row. This is also what many sink connectors expect, and the Confluent JDBC Sink Connector is no different:

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+[source,json,indent=0]
+----
+{
+	"field1" : "newvalue1",
+	"field2" : "newvalue2"
+}
+----
+
+Debezium provides a bridge between those two formats in a form of a https://cwiki.apache.org/confluence/display/KAFKA/KIP-66%3A+Single+Message+Transforms+for+Kafka+Connect[single message transform].
```
rhauch: > Although we think it's actually a great thing that Debezium CDC connectors provide as much detail as possible, we also make it easy for you to transform Debezium's "envelope" format into the "row" format that is more widely-used by connectors. Debezium provides ...

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+The `UnwrapFromEnvelope` transformation automatically extracts a new row record and thus effectively _flattens_ the complex record into a simple one consumable by other connectors.
+The SMT is applicable at both source and sink connectors but we generally recommend to apply it at sink connectors as then the records stored in Kafka will contain an associated context.
```
rhauch: > You can use this SMT on the source connector to transform the message _before_ it is written to Kafka, or you can instead store the source connector's richer envelope form of the message in Kafka and use this SMT on the sink connector to transform the message _after_ it is read from Kafka and before it is passed to the sink connector. Both options work, and it just depends on whether you find the envelope form of the message useful for other purposes.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+In our example we apply the SMT at sink connector using configuration properties
+
+[source,indent=0]
+----
+        "transforms": "unwrap", 
+        "transforms.unwrap.type": "io.debezium.transforms.UnwrapFromEnvelope", 
+----
+
+=== Delete records
+When a record is deleted from source database Debezium creates two records: delete record and tombstone record (for compaction).
```
rhauch: > When the Debezium connector detects a row is deleted, it creates two event messages: a delete event and a tombstone message. The delete message has an envelope with the state of the deleted row in the `before` field, and an `after` field that is null. The tombstone message contains same key as the delete message, but the entire message value is null, and Kafka's log compaction utilizes this to know that it can remove any earlier messages with the same key.
> A number of sink connectors, including the Confluent's JDBC Sink Connector, are not expecting these messages and will instead fail if it sees either kind of message. The `UnwrapFromEnvelope` SMT will by default filter out both delete and tombstone records, though you can change this if you're using the SMT and want to keep one or both of these kinds of messages.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+Kafka Connect JDBC Sink is not able to work with neither of those to and any attempt to use them leads to `NullPointerException`.
+`UnwrapFromEnvelope` by default filters out both delete and tombstone records and no action is requested on users side.
+If a downstream connector is able to handle those events it is possible to configure `UnwrapFromEnvelope` to pass them downstream.
+
+=== Topic naming
+Last but not least there is a difference in naming of topics.
+Debezium uses fully qualified naming for target topics representing each table it manages.
+The naming follows the pattern `<instance-name>,<database-name>.<table-name>`.
+Kafka Connect JDBC Connector works with simple names `<table-name>`.
```
rhauch: "The Confluent JDBC Sink Connector ..."

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+In complex scenarios the user would deploy https://kafka.apache.org/documentation/streams/[Kafka Streams] framework to establish elaborated routing between source and target routes.
```
rhauch: > In a more complex pipeline use case, we might use https://kafka.apache.org/documentation/streams/[Kafka Streams] to perform a more complex transformation, aggregation, and/or join of multiple topics, and this same Streams application could also route the messages to the correct topic.
>
> But in our simpler example we can instead use Kafka Connect's built-in `RegexRouter` SMT to send the messages to a different topic.

@jpechane, might be useful to say whether this SMT will be used in the source or sink connector. IIUC, you are using it in the source connector, and if that's right you might add something like this to the previous paragraph: "Again, we could use this SMT in either the source or sink connectors, but for this example we're going to use it in the source connector so we can choose the names of the Kafka topics where the records will be written."

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+1. unwrapping Debezium complex format into a simple one
+2. automatically create target tables
+3. insert a row if it does not exist or update an existing one
+4. identify primary key stored in Kafka's record value field
+
+[source,bash,indent=0]
+----
+curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @source.json
+----
+The registration request contains SMT for routing records into topics with simple name.
```
rhauch: How about showing the source connector configuration like you did for the sink connector, and explaining some of the configuration parameters?

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+docker-compose exec postgres bash -c 'psql -U $POSTGRES_USER $POSTGRES_DB -c "select * from customers"'
+ last_name |  id  | first_name |         email         
+-----------+------+------------+-----------------------
+ Thomas    | 1001 | Sally      | sally.thomas@acme.com
+ Bailey    | 1002 | George     | gbailey@foobar.com
+ Walker    | 1003 | Edward     | ed@walker.com
+ Kretchmar | 1004 | Anne       | annek@noanswer.org
+----
+
+Next we try to add a new record and verify that it was replicated into target database.
```
rhauch: I think it's beneficial to state once again that we're going from MySQL to PostgreSQL, so I'd suggest changing this sentence:

> With the connectors still running, we can add a new row to the MySQL database and then check that it was replicated into the PostgreSQL database.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+docker-compose exec -postgres bash -c 'psql -U $POSTGRES_USER $POSTGRES_DB -c "select * from customers"'
+ last_name |  id  | first_name |         email         
+-----------+------+------------+-----------------------
+...
+Doe        | 1005 | John       | john.doe@example.com
+(5 rows)
+----
+
+You can further play with the master data in MySQL and add a new data or modify an existing one and you can check the impact of the operations on target database.
+
```
rhauch: How about a "Summary" section that simply states what we were able to accomplish, something like:

> We set up a simple streaming data pipeline to replicate data in near real-time from a MySQL database to a PostgreSQL database. We accomplished this using Kafka Connect, the Debezium MySQL source connector, the Confluent JDBC sink connector, and a few SMTs -- all without having to write any code. And since it is a streaming system, it will continue to capture all changes made to the MySQL database and replicating them in near real time.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
+= Streaming data to a downstream database
+Jiri Pechanec
+:awestruct-tags: [ mysql, postgres, smt, example ]
+:awestruct-layout: blog-post
+
+Recently there was a short https://www.confluent.io/blog/simplest-useful-kafka-connect-data-pipeline-world-thereabouts-part-1/[series] of blogposts on https://www.confluent.io/[Confluent's] blog.
```
gunnarmorling: How are these posts related to Debezium? As it stands, that sentence is a bit disconnected, esp. given it's the opener.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+In this blogpost we will try to create a simple scenario: we are going to have a MySQL database instance running as a master database that feeds a Kafka broker with events representing a source of truth.
+Next we will add PostgreSQL database that will listen one of the topics and will replicate the state of the MySQL instance to its own data.
```
gunnarmorling: "listen *to*"; also, technically, it's not PG itself listening but rather the sink connector.

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+To simplify the setup a little bit we will use only one Kafka Connect instance that will contain all connectors so the single instance will serve as an event producer and an event consumer.
+
+.A Simplified topology
+[#img-general]
+[caption="Figure 2: "]
+image::dbz-to-jdbc-simplified.svg[Scenario topology]
+
+== Configuration
+We will use this https://github.com/debezium/debezium-examples/tree/master/unwrap-smt[compose] for a fast deployment of the demo.
```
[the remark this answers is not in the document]
gunnarmorling: I think so, it's a complete example with Docker Compose file etc.

blog/2017-09-25-streaming-to-another-database.adoc
gunnarmorling: *The* Kafka...

blog/2017-09-25-streaming-to-another-database.adoc
gunnarmorling: Why is it btw. that the sink connector doesn't accept DELETEs, @rhauch? Seems like a useful addition?

blog/2017-09-25-streaming-to-another-database.adoc
```diff
@@ -0,0 +1,222 @@
...
+...
+Doe        | 1005 | John       | john.doe@example.com
+(5 rows)
+----
+
+You can further play with the master data in MySQL and add a new data or modify an existing one and you can check the impact of the operations on target database.
+
+== What's next?
+
+In a future blogpost we will reproduce the same scenario with ElasticSearch as a target for events.
```
gunnarmorling: ElasticSearch=Elasticsearch (I used to do it incorrectly myself but then was educated by a friend :)).

blog/2017-09-25-streaming-to-another-database.adoc
rhauch: I think it helps to understand the context in which each connector was developed. Many of the sink connectors, including Confluent's Elasticsearch sink connector, were intended to capture the messages in one or more topics and write them out to Elasticsearch, regardless of whether those topics are compacted or not. And since deletes and tombstones only make sense on _compacted topics_, there was no special consideration for these kinds of messages.

I think the same is true for Confluent's JDBC sink connector, although that also was developed in conjunction with Confluent's JDBC _source connector_. My guess is that one more factor for not considering deletes and tombstones in the sink connector is that the source connector is not able to even detect deleted rows.

Although its certainly possible for event-driven applications to produce events that represent a removal, it is not necessary for these to be a tombstone. Really, CDC-like connectors may be one of the few source connectors that care to generate not only delete messages to capture the before state of the rows but also tombstone messages to signal Kafka that any previous messages with the same key could be compacted/removed.

blog/2017-09-25-streaming-to-another-database.adoc
rhauch: It may make sense to swap the first and second paragraphs so that the first sentence or paragraph says what this post is going to show how to do. The second paragraph can then follow that up by tying it to some of the other recent blog posts. Something like:

> In this blogpost we will try to create a simple streaming data pipeline to continuously capture the changes in a MySQL database and replicate them in near real-time into a PostgreSQL database. We'll show how to do this without writing code by instead using and configuring Kafka Connect, the Debezium MySQL source connector, the Confluent JDBC sink connector, and a few single message transforms (SMTs).
>
> This approach of replicating data through Kafka is really useful on its own, but it becomes even more advantages when we can combine our near real-time streams of data changes with other streams, connectors, and stream processing applications. A recent https://www.confluent.io/blog/simplest-useful-kafka-connect-data-pipeline-world-thereabouts-part-1/[Confluent blog post series] shows a similar streaming data pipeline but using different connectors and SMTs. What's great about Kafka Connect is that you can mix and match connectors to move data between multiple systems.

blog/2017-09-25-streaming-to-another-database.adoc
rhauch: Actually, as I mention in my comments, it's _Confluent's_ connector.

blog/2017-09-25-streaming-to-another-database.adoc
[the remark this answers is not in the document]
jpechane: Yes, it is in our new exmples repo

blog/2017-09-25-streaming-to-another-database.adoc
jpechane: Let me disagree with this point - I've made a note in text above that the connector is developed by confluent but the official name is Kafka Connect JDBC Connector.
At least it is stated in GitHub repo https://github.com/confluentinc/kafka-connect-jdbc/ and that's what I tried to consistently use in the text.

jpechane:
Thanks @rhauch and @gunnarmorling for the great feedback. I've incorporated almost all of your comments.

blog/2017-09-25-streaming-to-another-database.adoc
gunnarmorling: I'm not so much concerned about the technicality of tombstones or compaction, but rather that it seems not possible to convey the information that a record has been deleted to a sink such as Elasticsearch. Is the idea to only ever work with logical deletes, i.e. setting some "deleted marker" which would cause records to not be shown in results?

gunnarmorling:
@jpechane This reads really great now! I've pushed one more commit to this PR with a few typo fixes and other small improvements. I'll squash everything into a single commit before merging.

My last question is should we show our own routing SMT instead of the normal regex SMT?

jpechane:
> My last question is should we show our own routing SMT instead of the normal regex SMT?

I think it is different use case and even if it would do the same (I cannot confirm this) that it would be abusing of it and can make a confusion.

gunnarmorling:
I wouldn't say it's abusing it, but indeed all its powers are not needed for this very case. Going to merge :)

PR #79, opened 2017-11-01
DBZ-421 Adding instructions on OpenShift deployment to website

gunnarmorling:
https://issues.jboss.org/browse/DBZ-421

@jpechane There you go :) Smoothened it a bit here and there, added new intro, removed Minishift parts, added a screenshot of the OS Web Console and example output from the topic.


jpechane:
Should not this https://github.com/gunnarmorling/debezium.github.io/blob/14efa0bc2e2af43ae707053edd8ef45458c410c5/_partials/leftcol-doc.html.haml#L7 line be also updated?

gunnarmorling:
> Should not this line be also updated?

I guess :) Force-pushed an update.

docs/openshift.asciidoc
```diff
@@ -0,0 +1,220 @@
...
+
+Next we will create a Kafka Connect image with the Debezium connectors installed and then deploy a Kafka broker cluster and a Kafka Connect cluster:
+
+[source%nowrap]
+----
+# Deploy a Kafka broker
+oc new-app -p ZOOKEEPER_NODE_COUNT=1 barnabas
+
+# Build a Debezium image
+export DEBEZIUM_VERSION=0.6.1
```
jpechane: Can we use a reference to `debezium-version`?

docs/openshift.asciidoc
```diff
@@ -0,0 +1,220 @@
...
+== Verifying the Deployment
+
+Next we are going to verify whether the deployment is correct by emulating the link:/docs/tutorial/[Debezium Tutorial] in the OpenShift environment.
+
+First we need to start a MySQL server instance containing some example tables:
+
+[source%nowrap,bash]
+----
+# Deploy pre-populated MySQL instance
+oc new-app --name=mysql debezium/example-mysql:0.6
```
jpechane: Can we use a reference to `debezium-docker-label`?

docs/openshift.asciidoc
```diff
@@ -0,0 +1,220 @@
+= Deploying Debezium on OpenShift
+:awestruct-layout: doc
+:linkattrs:
+:icons: font
+
+The following describes how to set up the Debezium connectors for change data capture on Red Hat's https://www.openshift.com/[OpenShift] container platform.
```
jpechane: The following **text** ?

gunnarmorling:
I don't think it's needed. "The following" is the subject here.

jpechane:
Nice work thanks! No issues other then those minor ones

gunnarmorling:
Ok, sweet. Though the work is largely your's of course :) All your remarks
should be addressed, so can you push it?

PR #80, opened 2017-11-02
Fix typo in docs

charlesverdad:


jpechane:
Thanks for spotting and fixing @charlesverdad

PR #81, opened 2017-11-02
Fixing incorrect variable names in Avro instructions

gunnarmorling:
Our image defines its [own environment variables](https://github.com/debezium/docker-images/blob/master/connect-base/0.6/docker-entrypoint.sh#L23-L26), overriding the existing `CONNECT_...` ones.

I'm not sure whether that's really a good idea, but as we started with it, we should also do it for the two registry URL ones which currently need to be given via the original variables. I'll send a PR for the images repo for that.


jpechane:
Thanks for fixing!

PR #83, opened 2017-11-10
Blog - Devoxx talk

jpechane:


gunnarmorling:
Great idea! I think we should done it down a little bit, though. Otherwise it's praising ourselves too much. Let's express it more neutrally and leave judgement to readers.

Also let's add a link to [the slides](https://speakerdeck.com/gunnarmorling/streaming-database-changes-with-debezium) on Speaker Deck or even embed their control?

gunnarmorling:
Rebased and applied. Added one commit to make the video nicely adjust to the screen size. Thanks, @jpechane!

PR #87, opened 2017-11-16
FAQ entry on using Debezium's Kafka image on Docker for Mac or Windows

gunnarmorling:


jpechane:
Looks good - can you please confirm that you can connect from Docker containers to the IP adress of your host? As the client can be something running in the container it is necessary to verify the route is valid.

gunnarmorling:
Yes, that worked for me. I could ping the host from within a container and
also I think Connect wouldn't be able to connect Kafka if that didn't work.

jpechane:
Merged, thanks!

PR #88, opened 2017-11-23
DBZ-482 Clarifying requirement of database history and schema change …

gunnarmorling:
…topics to have a single partition

https://issues.jboss.org/browse/DBZ-482

In fact it may not be strictly needed, as we explicitly specify 0 as the partition when writing to these topics. But then using partitioned topics adds any benefits, so it seems reasonable to recommend against it in the docs, also acting as a safety net should we get it wrong in the implementation.


jpechane:
Thanks

PR #93, opened 2017-12-07
DBZ-502 Blogpost with Elasticsearch

jpechane:
https://issues.jboss.org/browse/DBZ-502
Can be merged after https://github.com/debezium/debezium-examples/pull/12


blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+&nbsp; +
+
+.A Simplified topology
+[#img-general]
+[caption="Figure 2: "]
+image::dbz-to-multiple-simplified.svg[Scenario topology]
+
+== Configuration
+We will use this https://github.com/debezium/debezium-examples/tree/master/unwrap-smt[compose] for a fast deployment of the demo.
```
gunnarmorling: s/compose/Docker Compose file/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+* https://hub.docker.com/r/debezium/zookeeper/[Apache ZooKeeper]
+* https://hub.docker.com/r/debezium/kafka/[Apache Kafka]
+* An https://github.com/debezium/debezium-examples/tree/master/unwrap-smt/debezium-jdbc[enriched] Kafka Connect / Debezium https://hub.docker.com/r/debezium/connect/[image] with changes
+** PostgreSQL JDBC driver placed into `/kafka/libs` directory
+** https://docs.confluent.io/current/connect/connect-jdbc/docs/index.html[Kafka Connect JDBC Connector] (developed by https://www.confluent.io/[Confluent]) placed into `/kafka/connect/kafka-connect-jdbc` directory
+* Pre-populated MySQL used in our link:docs/tutorial[tutorial]
+* Empty PostgreSQL
+* Empty ElasticSearch
+
+We again use the `UnwrapFromEnvelope` transformation to automatically extract a new row record and thus effectively convert the Debezium complex record into a simple one consumable by other connectors. Both JDBC and Elasticsearch connectors expect the same format.
```
gunnarmorling: s/complex record/change data message/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+== Example
+Let's move directly to our exmaple as that's where the changes are visible.
```
gunnarmorling: s/exmaple/example/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+First of all we need to deploy all components.
+[source,bash,indent=0]
+----
+export DEBEZIUM_VERSION=0.6
+docker-compose up
+----
+
+When all components are started we are going to register the Elasticsearch Sink connector writing into the ElasticSearch instance.
```
gunnarmorling: s/ElasticSearch/Elasticsearch/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+            "key.ignore": "false",                                                           (3)
+            "type.name": "kafka-connect"                                                     (4)
+        }
+    }
+}
+----
+
+The request configures these options:
+
+1. unwrapping Debezium's complex format into a simple one
```
gunnarmorling: How about "Extracting only the new rows state from Debezium's change data message"

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+2. extracting `id` field from the key `struct`, then the same key is used for the source and both destinations.
```
gunnarmorling: I don't quite get this part; what happens if we have a different id type? Btw. I recon it's not long but rather long or so?

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+This is to overcome limitation of Elasticsearch connector that supports only `int`s and `string`s as keys.
+3. use key from the event instead of generating a synthetic one
+4. type under which the events will be registered in Elasticsearch
```
gunnarmorling: Wouldn't it make more sense to use something like "customer"?

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+          "last_name" : "Walker",
+          "email" : "ed@walker.com"
+        }
+      }
+    ]
+  }
+}
+----
+
+With the connectors still running, we can add a new row to the MySQL database and then check that it was replicated into both the PostgreSQL database and the Elasticsearch:
```
gunnarmorling: s/the Elasticsearch/Elasticsearch/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+          "last_name" : "Doe",
+          "email" : "john.doe@example.com"
+        }
+      }
+...
+----
+
+== Summary
+
+We set up a complex strewaming data pipeline to synchronize MySQL database with another database and also with Elasticsearch instance.
```
gunnarmorling: s/strewaming/streaming/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
...
+We managed to keep the same identifier accors all systems which allows us to correlate records accross the system as the whole.
```
gunnarmorling: s/accors/across/
s/accross/across/

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,248 @@
+= Streaming data to a downstream Elasticsearch
+jpechane
+:awestruct-tags: [ mysql, postgres, elasticsearch, smt, example ]
+:awestruct-layout: blog-post
+
+We have seen how to stream a change data to a downstream database link:/blog/2017/09/25/streaming-to-another-database[few weeks ago].
+
+In this blog post we will follow the same approach to stream the data to an https://www.elastic.co/[Elasticsearch] engine for potential further search processing. But to make the matter a little bit more interesting we will stream the data to both PostgreSQL database and Elasticsearch so we will optimize access to data both via SQL query language and a full-text search.
```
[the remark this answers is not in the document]
jpechane: I used engine because it is defined as such in Elasticserch webpage, but server is probably better in this context.

blog/2017-12-07-streaming-to-elasticsearch.adoc
jpechane: I tried to add more explanation. The problem is visible in https://github.com/confluentinc/kafka-connect-elasticsearch/blob/22e7172c1d7550b37654364de55e9691578e54c5/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java#L80-L89. Feel free to reword as needed.

blog/2017-12-07-streaming-to-elasticsearch.adoc
jpechane: Yes and no. I kept it intentionally generic. It make sens to set it to customer if you are lsitening only on customers topic. If you add orders topic then th reasoning breaks. I'd say that type here describes it is a change data and the actual type (the originating  table) is defined by index name.

gunnarmorling: How is the index name derived btw.? Is it based on the connector name? The reason I'm asking is that you can't mix multiple types in a single ES index in current versions. So if the index name is derived from the connector instance name, it'd mean that one would typically have to use one connector per table/topic to be routed to ES.

jpechane: If `topic.index.map` is not set then the name of tipic is used, which is as good as it can be for this case.

gunnarmorling: Ok, got it. The type name is obsolete essentially as of ES 6, but I reckon the sink connector currently still requires it.

> I'd say that type here describes it is a change data and the actual type (the originating table) is defined by index name.

That's not the meaning of "type" in the ES context. It really refers to one type of documents within an index, so you could e.g. have documents of type "b2b_customer" and "b2c_customer" within one index. The type mapping associated with such type describes the fields of documents of that type. Mixing multiple types within one index won't be supported going forward.

Bottom line, using "customer" is what makes most sense to me for the example.

gunnarmorling:
@jpechane I've pushed one more commit with a few small adjustments. I'd say let's post it tomorrow afternoon and share on Twitter. Thanks!

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,255 @@
...
+jpechane
+:awestruct-tags: [ mysql, postgres, elasticsearch, smt, example ]
+:awestruct-layout: blog-post
+
+We have seen how to set up a change data stream to a downstream database link:/blog/2017/09/25/streaming-to-another-database[few weeks ago].
+
+In this blog post we will follow the same approach to stream the data to an https://www.elastic.co/[Elasticsearch] server for potential further search processing. But to make the matter a little bit more interesting, we will stream the data to both, a PostgreSQL database and Elasticsearch, so we will optimize access to the data via the SQL query language as well as via full-text search.
+
+== Topology
+The general topology for this scenario is shown in the following picture:
```
rhauch: I'd talk about this as the data flow. Something like:

> Here's a diagram that shows how the data is flowing through our distributed system. First, the Debezium MySQL connector is continuously capturing the changes from the MySQL database, and sending the changes for each table to separate Kafka topics. Then, the Confluent JDBC sink connector is continuously reading those topics and writing the events into the PostgreSQL database. And, at the same time, the Confluent Elasticsearch connector is continuously reading those same topics and writing the events into the Elasticsearch system.

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,255 @@
...
+
+.A General topology
+[#img-general]
+[caption="Figure 1: "]
+image::dbz-to-multiple.svg[Scenario topology]
+
+&nbsp; +
+
+We are going to simplify the deployment again, so we will use only one Kafka Connect instance that will contain all connectors and will serve as both as entry point and exit point for Kafka.
```
rhauch: And then this as the deployed components. Maybe

> We're going to deploy these components into several different processes. In this example, we'll deploy all three connectors to a single Kafka Connect instance that will write to and read from Kafka on behalf of all of the connectors.

blog/2017-12-07-streaming-to-elasticsearch.adoc
```diff
@@ -0,0 +1,255 @@
...
+* https://hub.docker.com/r/debezium/zookeeper/[Apache ZooKeeper]
+* https://hub.docker.com/r/debezium/kafka/[Apache Kafka]
+* An https://github.com/debezium/debezium-examples/tree/master/unwrap-smt/debezium-jdbc[enriched] Kafka Connect / Debezium https://hub.docker.com/r/debezium/connect/[image] with changes
+** PostgreSQL JDBC driver placed into `/kafka/libs` directory
+** https://docs.confluent.io/current/connect/connect-jdbc/docs/index.html[Kafka Connect JDBC Connector] (developed by https://www.confluent.io/[Confluent]) placed into `/kafka/connect/kafka-connect-jdbc` directory
+* Pre-populated MySQL used in our link:docs/tutorial[tutorial]
+* Empty PostgreSQL
+* Empty ElasticSearch
+
+We again use the `UnwrapFromEnvelope` transformation to automatically extract a new row record and thus effectively convert the Debezium change data message into a simple one consumable by other connectors. Both JDBC and Elasticsearch connectors expect the same format.
```
rhauch: This (using "again") makes a bit less sense if someone hasn't read the previous blog post.

Also, it's not as clear as it could be why this is even needed. Somewhere, you should probably explain that the connectors were developed separately from each other and chose to represent the events a bit differently. Debezium uses a more complex event structure so that it captures all of the information available. This `UnwrapFromEnvelope` single message transformation (SMT) collapses this complex change event structure into the same row-based format expected by the two sink connectors.

rhauch:
Can I also make a few suggestions about the diagrams? The `dbz-to-multiple` diagram should have one object for each connector, not have a single Confluent logo to represent two connectors. It seems like this is trying to show data flow, but then not having separate objects for each sink connector complicates that understanding. Also, the large red arrows are really distracting. :-)

Also, the `dbz-to-multiple-simplified` diagram seems like its trying to explain which processes do what. Again, the red arrows are super distracting, and it might work out better to use a different notation (boxes for each process, with boxes inside for each component). Also, shouldn't ZK be on this?

gunnarmorling:
Agreeing with @rhauch on his comments regarding the images and the arrows. Perhaps you can re-use the drawings from my recent presentation? I've just shared the link to the sources with you, so you can extract the SVG sources of the Kafka + Connect + DBZ visualizations from it and re-use for the post.

jpechane:
Thanks for the comments, I tried to incroporate them into the text.
Rearding diagrams I tried to change them before I move to @gunnarmorling's format.

jpechane:
@rhauch ZooKeeper is intentinally omitted as it is completely unrelated to the scenarios described.

jpechane:
@gunnarmorling It would probably make sense to rename the file before the publication - to change the date to the real one

gunnarmorling:
> Rearding diagrams I tried to change them before I move to @gunnarmorling's format.

I think they look better now, but I'm still not a huge fan. Any chance you re-do them based on the ones I did? It's not a blocker, but I feel they look a bit more orderly.

gunnarmorling:
> It would probably make sense to rename the file before the publication

+1

gunnarmorling:
Applied a few more changes and merged. Great write-up, thanks a lot, @jpechane! And also thanks a lot to @rhauch for your excellent remarks! It should be live on debezium.io in a bit.

PR #94, opened 2017-12-12
DBZ-498 Unknown datatypes handling

jpechane:
https://issues.jboss.org/browse/DBZ-498


jpechane:
Good, thanks for the improvements!

PR #104, opened 2018-01-12
DBZ-368 Document capacity of default memory settings

jpechane:
https://issues.jboss.org/browse/DBZ-368


gunnarmorling:
Applied. Thanks, @jpechane!

PR #105, opened 2018-01-17
Updating installation instructions to discuss plugin path instead of …

gunnarmorling:
…class path


gunnarmorling:
@jpechane A quick update as per today's ML question.

jpechane:
@gunnarmorling Fine for me, thanks.

PR #108, opened 2018-01-18
DBZ-548 FAQ for DECIMAL decoding from binary representation

jpechane:
https://issues.jboss.org/browse/DBZ-548


gunnarmorling:
Applied, thanks!

PR #110, opened 2018-01-22
DBZ-443 Adding docs for schema_only_recovery

pgoranss:


gunnarmorling:
@pgoranss thanks, LGTM.

Pushed one more commit to add a note when it's safe to use this mode. Will merge after the 0.7.2 release.

pgoranss:
Good additions, thanks!

PR #111, opened 2018-01-23
DBZ-409 Adding docs for Mongo CDC Event Flattening SMT

sairam881990:


gunnarmorling:
Applied and expanded a bit. Thanks, @sairam881990!

PR #112, opened 2018-01-23
Use Strimzi 0.1.0  instead of master

scholzj:
The website should not be using files from the master branch, which might be changing. It should use the release instead. This PR changes the links to the files from the 0.1.0 release.


gunnarmorling:
Good idea, merged. Thanks, @scholzj!

PR #113, opened 2018-01-24
DBZ-541 Filter history for monitored tables only

jpechane:
https://issues.jboss.org/browse/DBZ-541


docs/connectors/mysql.asciidoc
```diff
@@ -1457,9 +1457,16 @@ The following _advanced_ configuration properties have good defaults that will w
 |`database.history.skip.unparseable.ddl`
 |`false`
 |Boolean value that specifies if connector should ignore malformed or unknown database statements or stop processing and let operator to fix the issue.
-The safe default is `true`.
+The safe default is `false`.
```
gunnarmorling: Agh, good catch!

PR #115, opened 2018-01-25
Adding 0.7.2 release announcement and project roadmap

gunnarmorling:


jpechane:
Blogpost is fine, just please add link to the issues describing the limitations. You have there one for MongoDB flattening but you mention others in text.

gunnarmorling:
Do you just mean shortcomings of the Mongo SMT, or yet something else?

jpechane:
Geospatial arrays too

jpechane:
A roadmap is fine but I think to would be cool to have a picture like this https://time.graphics/line/52162

gunnarmorling:
Interesting idea. It'd requires us to stick specific release dates to it,
though. Let's keep that for a follow-up :)

gunnarmorling:
Rebased and applied. Thanks for reviewing, @jpechane!

PR #116, opened 2018-01-25
DBZ-574 Panel with 7 latest blogposts

jpechane:
https://issues.jboss.org/browse/DBZ-574

### DIFF
--- a/_layouts/blog-post.html.haml
+++ b/_layouts/blog-post.html.haml
@@ -2,59 +2,63 @@
 layout: project
 ---
 
-.row
-  .col-md-8.col-md-offset-2
-    .text-right
-      %h3
-        Subscribe
-        %a.rss{:href => "/blog.atom"}
-          %i.icon-rss
-    = partial('post.html.haml', {"post" => page})
-    - post = page["post"]
-    - if page.author
-      - authors = page.author.split(",");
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Wow, that was quick :) The menu is cool, but there's quite some whitespace between the menu and the actual contents, making the column with the latter quite narrow. Any chance to have a bit less whitespace for the sake of more width for the contents?

jpechane:
@gunnarmorling There's always a chance :-). I tried my best so hopefully it is ok now

jpechane:
@gunnarmorling Anyfurther comments on this? Are the margins still too big?

gunnarmorling:
@jpechane Yeah, looks much better now. Could we only have the same amount of padding on the right side of the text as on the left side of the menu? I think then it'd be perfect. And it'd also need a rebase to current HEAD. Thanks!

gunnarmorling:
screen shot 2018-03-09 at 22 26 12

To clarify that's what I meant. The padding on the left side is great, would be nice to have the same on the right side, too.

jpechane:
@gunnarmorling Rebased and made padding on both sides

gunnarmorling:
Looks great! Thanks, @jpechane!

PR #117, opened 2018-01-26
Document timing issues when primary key is changed

jpechane:


jpechane:
@gunnarmorling Any further comments on this one?

gunnarmorling:
Thanks, @jpechane; applied. Sorry for the delay, this has fallen through the cracks a bit.

PR #125, opened 2018-02-15
Adding 0.7.3 release announcement

gunnarmorling:


gunnarmorling:
@jpechane here's the announcement. 

Bad luck with Maven Central :( WDYT, should we still push out everything (with a short sentence saying that the sync is pending), or rather wait until the artifacts are there?

PR #127, opened 2018-03-05
[DBZ-602] - update documentation

Crim:
Updated documentation for issue DBZ-602 (https://github.com/debezium/debezium/pull/456)

New configuration entry for snapshot.locking.mode:
image

Deprecation notice for snapshot.mininal.locks:

image


gunnarmorling:
LGTM, thanks @Crim. Can you just add a sentence on "NONE" that it's only safe to be used if no schema changes are happening while the snapshot is taken? I've added the same to the enum in the code in https://github.com/debezium/debezium/commit/f055020613385141564e29c360bc5c7dbe018ca0#diff-d565df219003c24920bee4de756b0324R798.

Crim:
I think I got what you requested!

gunnarmorling:
Rebased and applied. Great work here and with the actual code change, your contribution is much appreciated. Thanks a lot, @Crim!

PR #128, opened 2018-03-07
DBZ-611 String encoding for decimals

jpechane:


gunnarmorling:
Applied. Thanks, @jpechane!

PR #133, opened 2018-03-08
Blog post on creating DDD aggregates

gunnarmorling:
Hey @hpgrahsl, here's my update on the blog post. It's rebased to the current master and based on your original PR. I'll close the latter as it's superseded by this one.

I really like the part you've written, thanks a ton for that! I've applied only small changes (related to casing and minor things like that) to it and otherwise added the Docker bits, expanded the intro, outlook and some more. Happy to learn about your feedback on these parts :)

The last thing missing is the author info (and I need to look into how to have two authors instead of one) and the todo on the follow-up post. I think I forgot one part you had in mind for that.

Really looking forward to publishing this!

PR #134, opened 2018-03-08
ddd blog post additions

hpgrahsl:


gunnarmorling:
Thx, integrated the two additional commits into #133.

PR #135, opened 2018-03-08
Allowing multiple authors per post

gunnarmorling:
@jpechane A quick change for the upcoming joint post with Hans-Peter.

### DIFF
--- a/_layouts/blog-post.html.haml
+++ b/_layouts/blog-post.html.haml
@@ -12,45 +12,47 @@ layout: project
     = partial('post.html.haml', {"post" => page})
     - post = page["post"]
     - if page.author
-      - author = site.authors[page.author];
-      - if author
-        .well
-          .row
-            .col-md-8
-              %h2<= author.name;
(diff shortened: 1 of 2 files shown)


_layouts/blog-post.html.haml
```diff
@@ -12,45 +12,47 @@ layout: project
...
     - post = page["post"]
     - if page.author
-      - author = site.authors[page.author];
-      - if author
-        .well
-          .row
-            .col-md-8
-              %h2<= author.name;
-              - if author.key?(:role)
+      - authors = page.author.split(",");
```
jpechane: The change works well but here I'd suggest an improvement, apply `strip` on each author. The reason is that it is quite natural to write `author1, author2` (note the space after comma) which brekas the current code.

gunnarmorling: Yes, very good idea. Thanks! Will do that and merge.

gunnarmorling:
Applied.

PR #136, opened 2018-03-13
Upgrade to awestruct 0.5.7

jpechane:
The experience is infitely better after upgrade - the incremental rebuild actually works. I tried also 0.5.7 but we are bitten by https://github.com/awestruct/awestruct/issues/513 in this case


jpechane:
Do not merge yet, it seems that 0.5.6 is broken in different way :-(

gunnarmorling:
FWIW, in.relation.to is on 0.5.7 already. Perhaps they found a way to avoid
that issue you ran into?

jpechane:
@gunnarmorling I spoke to i.r.t guys and they told me that the issue is not present with ruby 2.3.1 - probably. They provided a bugfix but it was never released. I've asked for planned releases at awestruct project but got no reply so far.
We would need to prepare our own ruby image because of this as there is no `2.3.1` tag on ruby image. I made a hack/workaround that removes the offending line - it is just a logging line so no harm is done.

gunnarmorling:
Thanks! So what should we do, merge this PR and go to 0.5.7? I.e. did you
apply this work-around in this PR?

jpechane:
@gunnarmorling  Yes, it is in, check `Rakefile` changes. There are two changes as the location is different for locally used imge and for travis

gunnarmorling:
Rebased and applied. Thanks a lot, @jpechane, indeed the flow when doing page updates is much nicer now.

PR #140, opened 2018-03-27
DBZ-673 Updating type of txId in PG connector docs

gunnarmorling:
https://issues.jboss.org/browse/DBZ-673


gunnarmorling:
Merged into upstream 0.8.

PR #142, opened 2018-03-28
Document differences for MongoDB 3.6

jpechane:


docs/configuration/mongodb-event-flattening.asciidoc
```diff
@@ -16,6 +16,11 @@ Please see below for a descriptions of known limitations of this transformation.
 This SMT is supported only for the MongoDB connector.
 ====
 
+[CAUTION]
+====
+There is a https://issues.jboss.org/browse/DBZ-624[known issue] for update messages when MongoDB 3.6 is used.
```
gunnarmorling: I think we should be a bit more specific and state that the SMT won't work with 3.6 for the time being.

docs/connectors/mongodb.asciidoc
```diff
@@ -368,6 +368,15 @@ When we compare this to the value in the _insert_ event, we see a couple of diff
 * The `source` field structure has the same fields as before, but the values are different since this event is from a different position in the oplog.
 * The `ts_ms` shows the timestamp that Debezium processed this event.
 
+[WARNING]
+====
+The content of the `patch` field is provided by the MongoDB itself and its exact format depends on the version.
```
gunnarmorling: s/the MongoDB/MongoDB/

jpechane:
@gunnarmorling Comments incorporated

gunnarmorling:
Applied to upstream/0.8. Thanks!

PR #143, opened 2018-03-29
DBZ-676 Document charset limitations

jpechane:


docs/connectors/postgresql.asciidoc
```diff
@@ -36,6 +36,12 @@ Since this is a relatively new feature, it has some limitations which are also r
 . Logical Decoding replication slots are only supported on `master` servers: this means that when there is a cluster of PostgreSQL servers, the connector can only run on the active `master` server. It cannot run on `hot` or `warm` standby replicas. If the `master` server fails or is demoted, the connector will stop. Once the `master` has recovered the connector can simply be restarted. If a different PostgreSQL server has been promoted to `master`, the connector configuration must be adjusted before the connector is restarted. Make sure you read more about how the connector behaves link:#when-things-go-wrong[when things go wrong].
 ====
 
+[IMPORTANT]
+====
+Debezium currently supports only database with multibyte character encoding.
```
gunnarmorling: Isn't this a bit too broad? There are other MBCS than UTF.

jpechane:
Right, I narrowed to scope to 'UTF-8', probably only single supported unicode encoding - https://www.postgresql.org/docs/9.6/static/multibyte.html

gunnarmorling:
Thanks, @jpechane!

PR #146, opened 2018-04-17
Remove extra dash in -it command

eric-weaver:
Removes an extra dash `(with '--it')` -> `(with '-it')`


jpechane:
@eric-weaver Merged, thanks for sptting and fixing!

PR #147, opened 2018-04-18
DBZ-355 KSQL Demo with Debezium

jpechane:
https://issues.jboss.org/browse/DBZ-355


gunnarmorling:
Rebased and applied. Also renamed to current date and applied a few typo/language fixes. Great write-up overall, thanks a lot, @jpechane!

PR #151, opened 2018-05-31
[DBZ-706] Updated documentation for new include.query option.

Crim:


docs/connectors/mysql.asciidoc
```diff
@@ -629,7 +646,8 @@ Let's look at what a _create_ event value might look like for our `customers` ta
       "file": "mysql-bin.000003",
       "pos": 154,
       "row": 0,
-      "snapshot": true
+      "snapshot": null,
```
Crim: I modified the snapshot value here to be consistent with the following examples.  Not sure if that was what I should have done or not.

gunnarmorling: LGTM.

gunnarmorling:
Merged into upstream 0.8. Thanks!

PR #152, opened 2018-06-04
DBZ-191 Mentioning default values in MySQL connector docs

gunnarmorling:
@jpechane @sweat123 Here's a quick update to docs regarding the default value issue.

### DIFF
--- a/docs/connectors/mysql.asciidoc
+++ b/docs/connectors/mysql.asciidoc
@@ -275,7 +275,7 @@ Each message written to the schema change topic will have a message key that con
   {
     "schema": {
       "type": "struct",
-      "name": "io.debezium.connector.mysql.SchemaChangeKey"
+      "name": "io.debezium.connector.mysql.SchemaChangeKey",
       "optional": false,
       "fields": [
         {
@@ -298,7 +298,7 @@ Meanwhile, the schema change event message's value will contain a structure cont
   {
     "schema": {
       "type": "struct",
-      "name": "io.debezium.connector.mysql.SchemaChangeValue"
+      "name": "io.debezium.connector.mysql.SchemaChangeValue",
       "optional": false,
       "fields": [
         {
@@ -434,7 +434,7 @@ Every change event for the `customers` table while it has this definition will f
   {
     "schema": {
       "type": "struct",
-      "name": "mysql-server-1.inventory.customers.Key"
+      "name": "mysql-server-1.inventory.customers.Key",
       "optional": false,
       "fields": [
         {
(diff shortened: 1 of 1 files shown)


docs/connectors/mysql.asciidoc
```diff
@@ -917,6 +917,12 @@ Columns that store strings are defined in MySQL with a character set and collati
 
 Other data type mappings are described in the following sections.
 
+If present, a column's default value will be propagated to the corresponding field's Kafka Connect schema.
```
jpechane: @gunnarmorling IIRC this is true also for cases where default value is set to all zeroes.

gunnarmorling:
> his is true also for cases where default value is set to all zeroes.

I'm not quite sure what you mean. Isn't this a form of "If present"?

jpechane:
@gunnarmorling Checkt this https://github.com/debezium/debezium/blob/master/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValuePreConverter.java#L93

gunnarmorling:
Hum, hum. Why is this even the case to begin with? I.e. why isn't that
value propagated as is?

jpechane:
@gunnarmorling All zeroes are invalid date/time but allowed by MySQL as a sort of dummy date - see https://dev.mysql.com/doc/refman/8.0/en/date-and-time-types.html We were thinking about mapping it to `NULL` but the problem would be with `NOT NULL` columns. And as it is not a valid value it cannot be propagated as is.

gunnarmorling:
@jpechane, ok turns out it already says this (added two years ago):

> MySQL allows zero-values for DATE, DATETIME, and TIMESTAMP columns, which are sometimes preferred over null values. These values cannot be represented using any of the Java types with either of the time.precision.mode options, and therefore the MySQL connector will represent them as null values when the column definition allows nulls, or as the epoch day when the column does not allow nulls.

So I think we're good? Also, has actually anything changed over that already described behavior with DBZ-191?

jpechane:
@jpechane @sweat123 I am afraid we are not good. I 've missed this part and obviously we are now violating this contract. We use epoch always but we should in fact use epoch or null depending on constraints.

sweat123:
@jpechane em. You are right! If the column definition allows null and the default value is all zero , we need set default value optional;

gunnarmorling:
@jpechane So I think we can merge this PR now as is; with @sweat123's latest code change it should match what's described in the docs again?

jpechane:
@gunnarmorling right

gunnarmorling:
Excellent; merged into upstream/0.8.

PR #155, opened 2018-06-07
DBZ-649 Support for heterogenous arrays

jpechane:
https://issues.jboss.org/browse/DBZ-649


docs/configuration/mongodb-event-flattening.asciidoc
```diff
@@ -67,6 +67,64 @@ transforms=unwrap,...
...
+Value `struct` will convert the array into a *struct* of *structs* in the similar way as done by http://bsonspec.org/[BSON serialization].
+The main *struct* contains fields named `_0`, `_1`, `_2` etc. where the name represents the index of the element in the array.
+Every element is then passed as the value for the give field.
+
+Let's suppose an example source MongoDB document with array with heterogeneous types
+[source,json,indent=0]
+----
+{
+    "_id": 1,
+    "a1": [
```
gunnarmorling: Why isn't this simply `"a1" : [ 1, "c" ]`?

jpechane: I wanted to demonstrate that i works not only for simple types but for documents too.

gunnarmorling:
Ah, I see. In that case I'd suggest to also add at least one more field to
the docs in the arrays, so to better motivate usage of a document as
opposed to single values.

jpechane:
@gunnarmorling Added one more field

PR #158, opened 2018-06-13
DBZ-728 Fix inconsistent names for history consumer/producer options

jpechane:
https://issues.jboss.org/browse/DBZ-728


gunnarmorling:
Good catch. Can you file an issue though and reference it from the commit message? Seems worth being mentioned in the change logs. Thx!

gunnarmorling:
Merged. Thanks!

PR #163, opened 2018-06-20
DBZ-746 add info about valid postgres slot names

kerdener:
add more information about valid names and link to Postgres documentation

https://issues.jboss.org/browse/DBZ-746


kerdener:
updated. thanks!

gunnarmorling:
Applied; thanks, @kerdener! FWIW, the commit author still isn't linked, but it's not much of a problem after all. You can still add the commit's e-mail address to your GH account if you like, in which case any commits with that e-mail will be attributed to your account.

PR #178, opened 2018-07-12
Debezium 0.8.0.Final release announcement

gunnarmorling:


jpechane:
@gunnarmorling LGTM, maybe just roadmpa should get updated too

gunnarmorling:
Yes, that's true. I'll try and spend some time on it next week.

PR #190, opened 2018-07-27
Improve Oracle datatype table formatting

jpechane:


gunnarmorling:
Applied; thanks, @jpechane!

PR #192, opened 2018-08-21
DBZ-866 Updating docs of 'source' block for MySQL connector

gunnarmorling:
@jpechane Small fall-out of reviewing https://github.com/debezium/debezium/pull/607.


jpechane:
@gunnarmorling Applied, thanks!

PR #194, opened 2018-08-29
DBZ-882 Correct whitelist and blacklist description

kleewho:
schemaName.tableName is not working for me, but I figured that it should be
databaseName.schemaName.tableName. If that's not the desired behavior you can
treat this as a bug report. If that is the desired behavior you have already a
change in the docs.

Please check https://issues.jboss.org/browse/DBZ-882


docs/connectors/sqlserver.asciidoc
```diff
@@ -410,11 +410,11 @@ The following configuration properties are _required_ unless a default value is
 
 |`table.whitelist`
 |
-|An optional comma-separated list of regular expressions that match fully-qualified table identifiers for tables to be monitored; any table not included in the whitelist will be excluded from monitoring. Each identifier is of the form _schemaName_._tableName_. By default the connector will monitor every non-system table in each monitored schema. May not be used with `table.blacklist`.
+|An optional comma-separated list of regular expressions that match fully-qualified table identifiers for tables to be monitored; any table not included in the whitelist will be excluded from monitoring. Each identifier is of the form _tableName_._schemaName_._tableName_. By default the connector will monitor every non-system table in each monitored schema. May not be used with `table.blacklist`.
```
krzwalko: According to https://github.com/debezium/debezium/blob/master/debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseConnectorConfig.java it should be probably:

_databaseName_._schemaName_._tableName_

kleewho: Yeah, thanks. Fixed

gunnarmorling:
@jpechane Any thoughts on ^?

jpechane:
@gunnarmorling SQL Server connector instance could theoretically monitor multiple databases in parallel but I am afraid it woould significantly complicate streaming logic so I'd keep the current limitation. I'd look at the filter common code if we can omit the database component from the name.

PR #204, opened 2018-09-22
A guide about how to install the wal2json logical decoding plugin at Postgresql

jpsoroulas:
A guide about how to install the wal2json logical decoding plugin at Postgresql

Details can be found at the discussion with @gunnarmorling :
https://groups.google.com/forum/#!topic/debezium/h5B52YTkGR4

The contents of the https://debezium.io/docs/connectors/postgresql/#output-plugin
have been merged into the section one of https://gist.github.com/jpsoroulas/30e9537138ca62a79fe261cff7ceb716
but NOT REMOVED from the original page (the debezium page stills intact).

No link for this page is created at side-bar menu.
Access the page at /docs/install/postgres-plugins/

### DIFF
--- a/docs/install/postgres-plugins.asciidoc
+++ b/docs/install/postgres-plugins.asciidoc
@@ -1,0 +1,404 @@
+= Postgresql, logical decoding output plugin installation
+:awestruct-layout: doc
+:toc:
+:toc-placement: macro
+:linkattrs:
+:icons: font
+:source-highlighter: highlight.js
+
+toc::[]
+
+At this document, the installation procedure of the https://github.com/eulerto/wal2json[wal2json] logical decoding output plugin
+and the https://www.postgresql.org/[PostgreSQL] database setup are performed. The installation and the tests are performed at the following environment/configuration:
+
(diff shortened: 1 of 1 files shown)


jpechane:
@jpsoroulas Thanks for the contribution, the text looks fine to me.
@gunnarmorling I think the page should be linked from somewhere, I do not like the notion of a hidden guide.

I am not 100 % sure of the text duplication, would not link be just enough? I understand the positives of have a single coherent text but I am not sure we would be able to keep it in sync.

gunnarmorling:
Rebased and applied. Thanks a lot, @jpsoroulas, and sorry again for the long delay with merging this.

I've also added a link to it from the actual Postgres connector config page. Regarding the redundancies between the two pages I think it would be nice to review the main connector page and get rid of the sections which are on that new page now.

jpsoroulas:
It is ok @gunnarmorling, it doesn't matter! 
I really understand the pressure from supporting many projects. I really like this project and I constantly keep up with the project's latest news and I watch the forum closely and actively. I also have some ideas for writing code examples, I will post them to you when I am ready.

Regarding the redundancies between the two pages, I'll see what I can do. 
And don't forget the PR #206 :+1:

PR #211, opened 2018-10-04
watch-topic needs kafka linked

jpechane:


gunnarmorling:
Are similar changes in order within the _README.md_ files in the examples repo?

jpechane:
We do not use `watch-topic` in examples.

gunnarmorling:
Ok, thanks for checking!

PR #212, opened 2018-10-11
commented on KSQL null unsupported printing for CDC topics

desavera:
Dear Jiri,

a really simple NOTE but I consider it useful.

### DIFF
--- a/docs/connectors/sqlserver.asciidoc
+++ b/docs/connectors/sqlserver.asciidoc
@@ -193,6 +193,10 @@ Change messages will contain the field's default value
 (unless an explicit column value had been given), so there should rarely be the need to obtain the default value from the schema.
 Passing the default value helps though with satisfying the compatibility rules when link:/docs/configuration/avro/[using Avro] as serialization format together with the Confluent schema registry.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


docs/connectors/sqlserver.asciidoc
```diff
@@ -193,6 +193,10 @@ Change messages will contain the field's default value
 (unless an explicit column value had been given), so there should rarely be the need to obtain the default value from the schema.
 Passing the default value helps though with satisfying the compatibility rules when link:/docs/configuration/avro/[using Avro] as serialization format together with the Confluent schema registry.
 
+[NOTE]
```
gunnarmorling: Thanks for your efforts, @desavera! It's not specific to the SQL Server connector, though, so I'd rather add it in form of an entry on the FAQ page.

So just to make sure I get it right, is this to say that KSQL won't select tombstone events by default? I'm confused by the "print messages" part.

/CC @jpechane, I think you folks discussed this.

jpechane: @desavera Yes exactly, I'like to see it as a question and answer on page https://debezium.io/docs/faq/

gunnarmorling:
Closing this, as I've added a more general item on this to the FAQ. Thanks for your efforts, @desavera.

PR #217, opened 2018-10-27
DBZ-953 Ability to configure decimal.handling.mode

pradeepmvn:
Docs to detail the options for decimal.handling.mode and mention the conversations. This is very similar to mysql docs.


gunnarmorling:
LGTM, thanks a lot, @pradeepmvn! We'll merge it once the next 0.9.x release is out.

PR #219, opened 2018-11-12
DBZ-925 Change `array.encoding` to correct options

renatomefi:
As in https://github.com/debezium/debezium/pull/672
Both documentation were wrong pointing to a value `struct` when it should've been `document`


gunnarmorling:
Merged. Thanks a lot, @renatomefi!

PR #220, opened 2018-11-12
Remove references to RDS using older version of wal2json

ksindi:
The caveats about wal2json being out of date in RDS can be removed since running a `git blame` shows those parts of the README.md are dated on Dec 20, 2017 and Mar 14, 2018, resp. That said, AWS RDS 9.6.10, 10.5 and 11 are all using wal2json commit hash 01c5c1ec which is dated Apr 29, 2018. I assume it's fine now. We have been testing postgres RDS and have not come across issues yet setting up Debezium.

Reference:
- AWS documention: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.version105
- wal2json version 01c5c1ec https://github.com/eulerto/wal2json/commit/01c5c1ec37b2c27baf00649c028b03936427596a

### DIFF
--- a/docs/connectors/postgresql.asciidoc
+++ b/docs/connectors/postgresql.asciidoc
@@ -62,12 +62,10 @@ It is possible to monitor PostgreSQL database running in https://aws.amazon.com/
 
 * The instance parameter `rds.logical_replication` is set to `1`.
 * Verify that `wal_level` parameter is set to `logical`; this might not be the case in multi-zone replication setups.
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Ah, great catch, @ksindi!

@jpechane, could you check whether we can remove any of the workarounds we put in place due to that?

ksindi:
Let me know if I should caveat the RDS postgres versions wal2json is updated on.

gunnarmorling:
Thanks a lot, @ksindi (and sorry for the delay, this one somewhat fell through the cracks)! I've expanded the information a bit, as the older wal2json versions are still used on older PG versions on RDS. That's clarified now.

gunnarmorling:
Applied. Thanks again, @ksindi!

PR #227, opened 2018-11-23
DBZ-841 Avoiding HTTPS -> HTTP -> HTTPS redirects;

gunnarmorling:
URLs like "/docs" will be redirected to "/docs/", but going through HTTP once.

Hey @jpechane, this is a left-over from DBZ-841. I replaced (hopefully) all URLs in the menus without trailing "/". So that's a big improvement already.

But there are still quite a lot of links without "/" left in many blog posts. I found them using this regex:

    link:/.*?[^\/]\[.*?\]

The problem is that I don't know how to replace them properly, because some of them refer to an anchor (e.g. `link:/docs/connectors/mysql#configuration[configuration file for the MySQL Connector]`), in which case the "/" must be added before the anchor and not at the very end. Perhaps you have an idea for fixing those, too?

In any case, the menu links are much more important as they are clicked much more often.


jpechane:
@gunnarmorling I've added further URL fixes

gunnarmorling:
Thanks, @jpechane. Somehow missed your update, cool you found a way to make this happen. Going to merge it.

PR #229, opened 2018-11-29
Update tutorial.asciidoc to note curl discrepancy on Windows

DaveSlinn:
Apparently you need to escape the double quotes in the json body of the request on Windows.


gunnarmorling:
Indeed haven't tried this on Windows, so thanks for spotting and fixing, @DaveSlinn.

PR #234, opened 2018-12-13
DBZ-966 Describe snapshot.delay.ms property

grzegorz8:
The property is now exposed in all connectors. Previously it was
available only for MySQL.


grzegorz8:
I basically copied the description including a note below which I do not fully understand, thus, I am not sure it need to be copied.

```
_Note: The connector may establish JDBC connections at its own discretion, so this should typically be used for configuration of session parameters only, but not for executing DML statements._
```

gunnarmorling:
> I basically copied the description including a note below which I do not fully understand, thus, I am not sure it need to be copied.

Ah, this one is wrong in the MySQL docs actually. It applies to the option above (database.initial.statements). Can you remove it from this PR? I'll adjust it in the MySQL docs.

grzegorz8:
@gunnarmorling Done

PR #238, opened 2019-01-11
[DBZ-1073] update website with type scale parameter propagation

amitsela:
Updates the website with the addition of the propagation of column type scale

### DIFF
--- a/docs/connectors/mysql.asciidoc
+++ b/docs/connectors/mysql.asciidoc
@@ -1563,7 +1563,7 @@ The following configuration properties are _required_ unless a default value is 
 
 |`column.propagate.source.type` 0.8.0 and later
 |_n/a_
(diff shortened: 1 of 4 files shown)


amitsela:
@gunnarmorling I see there's also https://github.com/debezium/debezium.github.io/blob/193519035f3f4d5404e200bae2dfa1331f272fcf/docs/connectors/oracle.asciidoc but I don't recall noticing the implementation when I added scale to mysql/postgresql, is that in the docs for oracle by mistake? or did I miss adding that in https://github.com/debezium/debezium/pull/726 ?

gunnarmorling:
LGTM, thanks! Will merge once 0.9.0.CR1 is out.

gunnarmorling:
Applied. Thanks a lot, @amitsela!

PR #239, opened 2019-01-13
Add missing `-f` flag to `kafkacat` cmd in the FAQ

kppullin:
In the FAQ, under the `How to change the offsets of the source database` section, the `kafkacat` command is missing the `f` flag character preceding the format string (it only contains `-` instead of `-f`).


gunnarmorling:
Thanls, @kppullin !

PR #253, opened 2019-02-04
SQL Server doc update

gunnarmorling:


gunnarmorling:
Reworked to remove the "incubating" note.

PR #258, opened 2019-02-15
Blog post on the outbox pattern

gunnarmorling:
Accompanying source code: https://github.com/debezium/debezium-examples/tree/master/outbox

~~@jpechane, I'm still going to refine this one, but here's the first draft. Any feedback on the overall flow would be highly welcomed.~~

Update: pretty happy with it now :)


blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+e.g. it could invoke some https://en.wikipedia.org/wiki/Representational_state_transfer[REST], https://grpc.io/[grpc] or other (synchronous) API provided by these services.
+This might create some undesired coupling, though: the sending service must know which other services to invoke,
+and it also must be prepared for these services temporarily not being available.
+Service meshes such as https://istio.io/[Istio] come in helpful here, by providing capabilities like circuit breakers,
+request routing and much more.
+
+One general issue of such a synchronous approach is that it lacks re-playability,
+i.e. the possibility for new consumers to arrive after events have been sent and still be able to consume the entire event stream from the beginning.
+
+This can be achieved by using an asynchronous data exchange approach:
```
jpechane: I think this is not fully correct - this implies that `asynchronous data exchange` == `replayabilty` but those two concepts overlap only partially.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+by having the order service propagate events through a durable message log such as http://kafka.apache.org/[Apache Kafka],
+new consumers can be added as needed, enabling use cases you might not have had in mind originally,
+and without touching the source service.
+E.g. consider a data warehouse which should keep information about all the orders ever placed, or some full-text search functionality on orders based on https://www.elastic.co/products/elasticsearch[Elasticsearch].
+Once the purchase order events are in a Kafka topic
+(and depending on the topic's retention policy, they could remain their for an arbitrary long time or even indefinitely),
```
jpechane: Should be `there` not `their`

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+new consumers can subscribe, process the topic from the very beginning and materialize a view of all the data in a microservice's database, search index, data warehouse etc.
+
+== The Issue of Dual Writes
+
+In order to provide their functionality, data-centered microservices will typically have their own local database.
```
jpechane: `local database` or `local data store`?

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+The reason being that we cannot have one shared transaction that would span the service's database as well as Apache Kafka.
+So in unfortunate circumstances it might happen that we end up with having the new purchase order persisted in the local database,
+but not having sent the corresponding message to Kafka
+(e.g. due to some networking issue).
+Or, the other way around, we might have sent the message to Kafka but failed to persist the purchase order in the local database.
+Both situations are undesirable;
+this may cause no shipment to be created for a seemingly successfully placed order.
+Or a shipment gets created, but then there'd be no trace about the corresponding purchase order in the order service itself.
+
+So how can this situation be avoided?
```
jpechane: I think a short mention of XA and why it was not chosen should be added.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+----
+Headers: eventId=d03dfb18-8af8-464d-890b-09eb8b2dbbdd
+Key: "4"
+Value: {"eventType":"OrderCreated","ts_ms":1550307598558,"payload":"{\"id\": 4, \"lineItems\": [{\"id\": 7, \"item\": \"Debezium in Action\", \"status\": \"ENTERED\", \"quantity\": 2, \"totalPrice\": 39.98}, {\"id\": 8, \"item\": \"Debezium for Dummies\", \"status\": \"ENTERED\", \"quantity\": 1, \"totalPrice\": 29.99}], \"orderDate\": \"2019-01-31T12:13:01\", \"customerId\": 123}"}
+Headers: eventId=49f89ea0-b344-421f-b66f-c635d212f72c
+Key: "4"
+Value: {"eventType":"OrderLineUpdated","ts_ms":1550308226963,"payload":"{\"orderId\": 4, \"newStatus\": \"CANCELLED\", \"oldStatus\": \"ENTERED\", \"orderLineId\": 7}"}
+----
+
+The `payload` field with the message values is the string-ified JSON representation of the original events.
```
jpechane: The article does not explain how the sringification was achieved.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,598 @@
...
+id            | uuid                   | not null
+aggregatetype | character varying(255) | not null
+aggregateid   | character varying(255) | not null
+type          | character varying(255) | not null
+payload       | jsonb                  | not null
+----
+
+Its columns are these:
+
+* `id`: unique id of each message; can be used by consumers to detect any duplicate events, e.g. when restarting to read messages after a failure.
```
jpechane: How is the `id` generated?

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Thx.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Both is fine I guess. But I like store better, too, changing it :)

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Ok, I'll try. But "not chosen" sounds as if it were an option; but AFAICS it isn't, as Kafka cannot be enlisted in an XA transcation.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Upon event instantiation. Clarified it.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Done.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Clarified it.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+This might be surprising at first.
+But it makes sense when remembering how log-based CDC works:
+it doesn't examine the actual contents of the table in the database, but instead it tails the append-only transaction log.
+The calls to `persist()` and `remove()` will create an `INSERT` and a `DELETE` entry in the log once the transaction commits.
+After that, Debezium will process these events (handling the `INSERT` and ignoring the `DELETE`).
+So we are able to capture the event added to the outbox table by means of CDC,
+but when looking at the contents of the table itself, it will always be empty.
+This means that no additional disk space is needed for the table
+(apart from the log file elements which will automatically be discarded at some point)
+and also no separate house-keeping process is required to stop it from growing indefinitely.
```
sherl0cks: Hmmm - I have very mixed feelings about this. This is related to the idea that events live in Kafka long term. Essentially it moves operational complexity from a well known industry space (relational DBs) towards a very poorly known industry space (kafka / k8s / ocp).

There are other complexities here too. I need to think about this.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+i.e. new consumers can be added as needed, enabling use cases you might not have had in mind originally,
+and without touching the source service.
+E.g. consider a data warehouse which should keep information about all the orders ever placed, or some full-text search functionality on purchase orders based on https://www.elastic.co/products/elasticsearch[Elasticsearch].
+Once the purchase order events are in a Kafka topic
+(and depending on the topic's retention policy, they could remain there for an arbitrary long time, or even indefinitely),
+new consumers can subscribe, process the topic from the very beginning and materialize a view of all the data in a microservice's database, search index, data warehouse etc.
+
+== The Issue of Dual Writes
+
+In order to provide their functionality, data-centered microservices will typically have their own local data store.
```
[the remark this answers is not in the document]
gunnarmorling: Interesting, @jpechane was stumbling upon it, too. I'll remove it.  I had it there to make a distinction to microservices that don't need a data store at all. But it's not too important.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+Simply issuing these two requests may lead to potential inconsistencies, though.
+The reason being that we cannot have one shared transaction that would span the service's database as well as Apache Kafka,
+as the latter doesn't support to be enlisted in a distributed (XA) transaction.
+So in unfortunate circumstances it might happen that we end up with having the new purchase order persisted in the local database,
+but not having sent the corresponding message to Kafka
+(e.g. due to some networking issue).
+Or, the other way around, we might have sent the message to Kafka but failed to persist the purchase order in the local database.
+Both situations are undesirable;
+this may cause no shipment to be created for a seemingly successfully placed order.
+Or a shipment gets created, but then there'd be no trace about the corresponding purchase order in the order service itself.
```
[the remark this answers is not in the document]
gunnarmorling: In which case could a re-ordering occur?

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+aggregateid   | character varying(255) | not null
+type          | character varying(255) | not null
+payload       | jsonb                  | not null
+----
+
+Its columns are these:
+
+* `id`: unique id of each message; can be used by consumers to detect any duplicate events, e.g. when restarting to read messages after a failure.
+Generated when creating a new event.
+* `aggregatetype`: the type of the aggregate root to which a given event is related; this could for instance be "purchase order" or "customer".
```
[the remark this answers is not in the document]
gunnarmorling: Good point, I'll add some explanatory sentence before that. That said, I don't see the term "aggregate" strictly tied to DDD, and mind you, last time I actually made the link, I got called out :) But in any case it's worth being explained upfront.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+using a representation specifically tailored towards its own needs.
+For instance, such view might be denormalized to efficiently support specific access patterns, or it may only contain a subset of the original data that's relevant to the consuming service.
+
+Durable logs also support re-playability,
+i.e. new consumers can be added as needed, enabling use cases you might not have had in mind originally,
+and without touching the source service.
+E.g. consider a data warehouse which should keep information about all the orders ever placed, or some full-text search functionality on purchase orders based on https://www.elastic.co/products/elasticsearch[Elasticsearch].
+Once the purchase order events are in a Kafka topic
+(and depending on the topic's retention policy, they could remain there for an arbitrary long time, or even indefinitely),
+new consumers can subscribe, process the topic from the very beginning and materialize a view of all the data in a microservice's database, search index, data warehouse etc.
```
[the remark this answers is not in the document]
gunnarmorling: I see; do you have an alternative for replayability? Also I'm curious, that thread you linked is nearly three years old, perhaps things have changed since then? I've seen references to long retention, but I'm sure what's the current agreed on thinking on that one is.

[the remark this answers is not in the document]
hpgrahsl: while it is right that very long (indefinite) storage brings its own challenges there are real-world examples where this is done. the "most famous" one that quickly comes to my mind is the NYT approach towards permanent storage https://www.confluent.io/blog/publishing-apache-kafka-new-york-times/ and a referring post https://www.confluent.io/blog/okay-store-data-apache-kafka/

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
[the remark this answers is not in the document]
hpgrahsl: @gunnarmorling given what happended last time in our joint article I'd suggest that if you really want to refer to the DDD idea of an aggregate you might want to reference Eric Evans?

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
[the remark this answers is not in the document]
sherl0cks: From the NYT blog:
```
Because the topic is single-partition, it needs to be stored on a single disk, due to the way Kafka stores partitions. This is not a problem for us in practice, since all our content is text produced by humans — our total corpus right now is less than 100GB, and disks are growing bigger faster than our journalists can write.
```
So that's 100GB for > 150 years of data. I think many, many use cases customers in production will have much greater different scale than  .66 GB / year and design systems to use more than one partition. 

I'm not saying that the approach is invalid, I'm just saying that proposing this sort of architecture (indefinite storage in the Kafka cluster, no event storage in the DB) has serious operational implication and at minimal we should highlight this fact so readers are aware of trade off & alternative.

As to reply ability, I think the most likely case is going to after an outage where we've used traditional message queueing for ages. The idea of a whole new service replaying 5 years of message history from multiple services to create an initial state is not a daily or weekly requirement at most organizations. If the requirement does come up, I would prefer to leverage an event log in my RDBMS and create a one off debezium connector for this specific purpose. This gets in to the whole idea of optionality I write about [here](https://medium.com/@justinmichaelholmes/creating-optionality-with-domain-event-logging-7eb8f1f422e7)

Again - these alternatives. Neither is right or wrong and all I am asking is that we highlight this approach as one of many.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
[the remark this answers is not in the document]
sherl0cks: The more I think about it, the more this is a corner case not worth exploring. Your previous points clearly communicate the challenge. Closing this comment.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
sherl0cks: Per the previous comment thread - I'm OK with this approach, but we should add a line here that mentions that this approach is a corollary of the indefinite storage in Kafka (which again is one option of several). I'm just worried that I'll get customers telling me "This is the one and only way to implement outbox! Gunnar said so!"

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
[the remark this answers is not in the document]
sherl0cks: @gunnarmorling if you don't want to address this as a DDD idea, thats 100% OK. But you should provide context to what you mean specifically. If you do want to address it as a DDD idea, then I think I know how to word it properly and I can send a PR if you would like it.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+which ensures that consumers of that topic will consume all the events related to one and the same aggregate root in the exact order as they were produced.
+* `type`: the type of event, e.g. "Order Created" or "Order Line Canceled". Allows consumers to trigger suitable event handlers.
+* `payload`: a JSON structure with the actual event contents, e.g. containing a purchase order, information about the purchaser, contained order lines, their price etc.
+
+=== Sending Events to the Outbox
+
+In order to "send" events to the outbox, code in the order service could in general just do an `INSERT` into the outbox table.
+However, it's a good idea to go for a slightly more abstract API, allowing to adjust implementation details of the outbox later on more easily, if needed.
+https://docs.jboss.org/weld/reference/latest/en-US/html/events.html[CDI events] come in very handy for this.
+They can be raised in the application code and will be processed _synchronously_ by the outbox event sender,
```
sherl0cks: May be worth mentioning that a Spring impl is quite similar. https://zoltanaltfatter.com/2017/06/09/publishing-domain-events-from-aggregate-roots/ and https://github.com/haithamshahin333/Domain-Event-Sample

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+By only modifying a single resource - the source service's own database -
+it avoids any potential inconsistencies of altering multiple resources at the same time which don't share one common transactional context
+(the database and Apache Kafka).
+By writing to the database first, the source service has instant "read your own writes" semantics,
+which is important for a consistent user experience, allowing query methods invoked following to a write to instantly reflect any data changes.
+
+At the same time, the pattern enables asynchronous event propagation to other microservices.
+Apache Kafka acts as a highly scalable and reliable backbone for the messaging amongst the services.
+Given the right topic retention settings, new consumers may come up long after an event has been originally produced,
+and build up their own local state based on the event history.
```
sherl0cks: This is again subject to the same "there are options here" qualification.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,701 @@
...
+i.e. consumers such as the shipping service may lag a bit behind producers such as the order service.
+Usually, that's just fine, though, and can be handled in terms of the application's business logic.
+For instance there'll typically be no need to create a shipment within the very same second as an order has been placed.
+Also, end-to-end delays of the overall solution are typically low (seconds or even sub-second range),
+thanks to log-based change data capture which allows for emission of events in near-realtime.
+
+One last thing to keep in mind is that the structure of the events exposed via the outbox should be considered a part of the emitting service's API.
+I.e. when needed, their structure should be adjusted carefully and with compatibility considerations in mind.
+This is to ensure to not accidentally break any consumers when upgrading the producing service.
+At the same time, consumers should be lenient when handling messages and for instance not fail when encountering unknown attributes within received events.
```
sherl0cks: This last point is really important, and its the key corollary to "messaging makes you more decoupled." It may be worth making this point earlier in the blog.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Added a reference to `@DomainEvents`.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
```diff
@@ -0,0 +1,697 @@
...
+----
+
+It's rather simple: for each event the CDI runtime will invoke the `onExportedEvent()` method.
+An instance of the `OutboxEvent` entity is persisted in the database -- and removed right away!
+
+This might be surprising at first.
+But it makes sense when remembering how log-based CDC works:
+it doesn't examine the actual contents of the table in the database, but instead it tails the append-only transaction log.
+The calls to `persist()` and `remove()` will create an `INSERT` and a `DELETE` entry in the log once the transaction commits.
+After that, Debezium will process these events (handling the `INSERT` and ignoring the `DELETE`).
```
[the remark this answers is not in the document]
gunnarmorling: Good idea; done.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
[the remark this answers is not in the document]
gunnarmorling: I've added some explanation now; @sherl0cks, do you think it's ok as such now? If not, a PR would be much appreciated.

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: I'm touching on it earlier on already (and expanded that one a bit):

> By explicitly emitting events via records in the outbox table,
the application code can work with events are structured in a way suitable for external consumers.
> This also ensures that event consumers won't break
when for instance altering the internal domain model or the `PurchaseOrder` table.

I think it's ok that way.

gunnarmorling:
Hey @sherl0cks, so first of all, thank you (and everyone else!) to all your comments, they are very much appreciated.

In particular the discussion around indefinite topic retention is a good one. This is presented in a slightly different angle now, please see the two note boxes "Dealing with Topic Growth" and "Deletion of Events from Kafka Topics" for the details. This now discusses compacting the event topics (so you'd only keep the last event pertaining to a given purchase order, customer etc., which requires that each event fully describes the represented aggregate, and not just the "delta") and also the idea of moving historic events to cold storage such as an S3 bucket. Could you perhaps review the last commit I've added and let me know whether you see any remaining issues? From my perspective, it's good to go and I hoped I've addressed all your remarks. Thanks again so much!

blog/2019-02-19-reliable-microservices-data-exchange-with-the-outbox-pattern.adoc
gunnarmorling: Should be addressed now.

gunnarmorling:
And applied; should be live in a few minutes. Thanks again everyone for all your comments, that helped a lot!

PR #259, opened 2019-02-18
DBZ-867 add property snapshot.mode for MongoDB

krizhan:
Added new property corresponding to the following commit:

https://github.com/debezium/debezium/pull/774

### DIFF
--- a/docs/connectors/mongodb.asciidoc
+++ b/docs/connectors/mongodb.asciidoc
@@ -622,6 +622,10 @@ The following configuration properties are _required_ unless a default value is 
 |_empty string_
 |An optional comma-separated list of regular expressions that match fully-qualified namespaces for MongoDB collections to be excluded from monitoring; any collection not included in the blacklist will be monitored. Each identifier is of the form _databaseName_._collectionName_. May not be used with `collection.whitelist`.
 
+|`snapshot.mode`
+|`initial`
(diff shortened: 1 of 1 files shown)


docs/connectors/mongodb.asciidoc
```diff
@@ -622,6 +622,10 @@ The following configuration properties are _required_ unless a default value is
 |_empty string_
 |An optional comma-separated list of regular expressions that match fully-qualified namespaces for MongoDB collections to be excluded from monitoring; any collection not included in the blacklist will be monitored. Each identifier is of the form _databaseName_._collectionName_. May not be used with `collection.whitelist`.
 
+|`snapshot.mode`
+|`initial`
+|Specifies the criteria for running a snapshot (eg. initial sync) upon startup of the connector. The default is *initial*, and specifies the connector reads a snapshot when a server is added to the replica set or when a task first obtains the current state of the replica set. The *never* option specifies that the connector should never use snapshots, instead the connector should proceed to tail the log.
```
gunnarmorling: > when a server is added to the replica set

This one is interesting, do you have a reference where this is happening? From looking it the code it seems rather it would do the snapshot if no earlier offset was committed. Also when taking a step back, simply adding a replica seems not like it'd warrant to take a new snapshot?

krizhan: Ohh yeahh. My apologies, I misread something beforehand. I'll fix it up.

gunnarmorling:
@jpechane, seems we missed this one during the 0.9.2 release. I'll merge it.

PR #265, opened 2019-03-23
Updated docs to use snapshot.isolation.mode

pradeepmvn:
Removed snapshot.locking.mode and added snapshot.isolation.mode to SQL Server Connector documentation


gunnarmorling:
Sorry for the delay, @pradeepmvn, this one fell through the cracks so far.

@jpechane, @grzegorz8, any objections?

jpechane:
@pradeepmvn Applied, thanks

PR #270, opened 2019-03-28
Update online-resources.asciidoc

olenasherbinin:


gunnarmorling:
Hi @olenasherbinin, thanks for sharing the link. It looks like an interesting solution and it's great to see that Debezium proves to be a useful component of it.

That being said, on the "online resources" list we try to gather links to community resources (talks, slide decks etc.) and related open-source solutions. So I'm not going to merge this PR, as it seems to not fall into these categories. Thanks for understanding!

PR #279, opened 2019-04-13
DBZ-1034 Added [ts/tstz/date]range type description to postgres docs

sciamp-dev:
Added documentations for postgres tsrange, tstzrange and daterange types.
Related to https://github.com/debezium/debezium/pull/830

Currently I added three rows in the first table. Dunno if it would be better to have a dedicated table/section for range types.


gunnarmorling:
@jpechane, seems this fell through the cracks in the 0.9.5 release. The functionality is merged, but the issue in JIRA still is open.

jpechane:
@gunnarmorling Yeah, the Jira issue does not have fix version assigned so the release process did not detect the limbo state.

PR #283, opened 2019-04-25
ADD: Index page with all blog posts so far

hashhar:
A page which lists all blog posts is needed so that people don't
have to click 10 times to see what all blog posts are there.


hashhar:
A preview of how it looks like is attached.

image

gunnarmorling:
Hi, @hashhar, thanks a lot! That's a great improvement. I'd have two suggestions for improvements:

* Don't add *all* the posts on a single page but add paging
* Show the first *n* words of each post on that overview page

Both can be seen live on the Hibernate blog: http://in.relation.to/

Would that be something you'd like to pick up? Btw. on which URL will this new page show up? debezium.io/blog?

I'm really cool to see these efforts, as I felt the website still is lacking in terms of discoverability of older contents, so kudos for kicking this off!

hashhar:
The pagination makes sense. I'll do that. The first n words parts I'll take a look into.

The current archive page lives at debezium.io/archive

hashhar:
image

We now show 10 posts per page on https://debezium.io/archive/.

The summary is constructed by taking out the first paragraph from the blog post.

gunnarmorling:
Yeah, that looks really cool! One thing I think would be good would be to have a "Read more..." button/link under each of the entries as on the in.relation.to blog. And I think we don't even need the /archive path, but just the main debezium.io/blog path should show be rendered in this way.

Thanks again, @hashhar, it's really a great improvement I had wished for for a long time :)

gunnarmorling:
Ah, and how does it control how much contents is shown of each post on the overview page? If possible, it'd be good to have explicit control (e.g. by means of a comment in the post files). Again we could see how it's done on the Hibernate blog. If it's doable, it'd be nice to make the choice explicit for all the posts showing up on the first page at least, so to avoid the very sparse announcement posts with just a single short sentence.

hashhar:
@gunnarmorling Currently it's just picking up the first paragraph.

We can also make it use the ASCII Doc preamble. ie. Everything from when the page starts to the first section.

```asciidoc
= Debezium 0.9.4.Final Released
gmorling
:awestruct-tags: [ releases, mysql, postgres, docker ]
:awestruct-layout: blog-post

It's my pleasure to announce the release of Debezium *0.9.4.Final*!

This is a drop-in replacement for earlier Debezium 0.9.x versions, containing mostly bug fixes and some improvements related to metrics.
Overall, https://issues.jboss.org/issues/?jql=project%20%3D%20DBZ%20AND%20fixVersion%20%3D%200.9.4.Final[17 issues] were resolved.

== MySQL Connector Improvements
```
In the above document everything before that `==` will be part of the preamble.

The best approach is to use the `==` for implicitly defining the summary, and then use a comment of the form `<!-- more -->` to explicitly define the summary.

hashhar:
I think this is done.

You can manually define a summary using `+++<!--more-->+++` anywhere in the asciidoc file.
If not defined, then we use the first section of the asciidoc as defined above.

image

hashhar:
To get the most out of this, you'll maybe need to add the marker or either define a section for earlier posts.

gunnarmorling:
> The best approach is to use the == for implicitly defining the summary, and then use a comment of the form  to explicitly define the summary.

Yes, that would be great!

gunnarmorling:
Hey @hashhar, finally getting back to this (sorry for the delay, I was on some conference travel for the last weeks). I just built your PR locally and it looks great. I'd have two more remarks:

* There should be no "Read More" Button if the full post is already shown on the main page
* Could we have a bit more margin at the top of the "Read More" Button; it seems a bit too close to the text before

Could you take a look into these two things? Thanks a lot!

hashhar:
@gunnarmorling I'll surely take a look into those over the weekend. A little swamped with work right now.

Good catch regarding the unnecessary READ MORE in some cases.

gunnarmorling:
Hey @hashhar, howdy? Do you still plan to wrap this up as per the last comments above? Or should someone else take over? Thanks!

hashhar:
@gunnarmorling Fixed and pushed. No more "READ MORE" when the entire content is present and also increased some padding.

hashhar:
@gunnarmorling let me know if this looks okay.

gunnarmorling:
@hashhar, thanks a lot this looks great! I'm only having issues with applying the explicit "read more" markers. Neither "" nor "++++++" seem to work. Any idea?

_ext/posts.rb
```diff
@@ -0,0 +1,137 @@
+require 'nokogiri'
+
+module Awestruct
+  module Extensions
```
gunnarmorling: Is this extension coming from somewhere else btw? Or is it written all from scratch?

_ext/posts.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      def extract_summary(post_content)
+        summary = post_content
+        # If the excerpt marker is present, we use it
+        if post_content.include?('<!--more-->')
+          summary = post_content.split('<!--more-->').take(1).join()
+        end
+
+        # Use preamble if present
+        preamble = Nokogiri::HTML::Document.parse(post_content).at("div#preamble")
```
gunnarmorling: This seems to be the problem why the excerpt markers don't work. The logic below should only be used when the marker isn't present. If I change it into that though, I'm getting weird rendering issues. I suppose it's caused by this working on HTML instead of the source asciidoc, so one ends up with an incomplete <div> or so. Have you actually tried this out?

gunnarmorling:
@hashhar, so I (mostly) made the marker work and have pushed a commit for this. It's based on [the work](https://github.com/hibernate/in.relation.to/blob/staging/_ext/posts.rb) from my colleagues in the Hibernate team.

There's just one last thing: it doesn't work for the two posts which have the "teaser" role (e.g. see blog/2019-03-14-debezium-meets-quarkus.adoc). Could you take a look at that one? Ideally, it'd just propagate the teaser paragraph as the excerpt, without the need for any explicit marker. Thanks!

hashhar:
I'll take a look tomorrow night and get back to you. Thanks for the additional commits.

gunnarmorling:
Hey @hashhar, any news on this question above? Thanks!

Naros:
I would like to add that I think it would also be useful to change the left navigation pane to render a tag cloud for all blogs rather than the _Latest Posts_.  It would allow users to navigate to very specific blogs of interest and we could re-use this brief blog highlight layout pagination there as well.

Naros:
@hashhar I just rebased your PR against the latest website repository and force pushed.  I'll take a look at this over the weekend and wrap it up.

Naros:
@gunnarmorling @hashhar, with the new website changes, I couldn't help myself.  So I pushed a few changes to address the blog list formatting and added a tag-cloud to the left navigation pane along with aligning the UI with the new website layout.

The following 2 open issues I will address sometime this weekend:

* Quarkus blog entry teaser problem that Gunnar mentioned above
* Replicate UI changes to the rendering of each individual blog entry

Naros:
Commit https://github.com/debezium/debezium.github.io/pull/283/commits/e858bbf41c724dd3961b9351de7f51ca00690ea6 resolves the last 2 remaining open issues, so this should be good so long as it continues to be fine with the team.

_ext/posts.rb
Naros: I didn't see a specific place now where this is a problem after properly handling the extraction of the summary from the teaser role.  If there are places this is still problematic, let me know.

_ext/posts.rb
Naros: It looks like its very similar to the one for in.relation.to.  I did make a few modifications to the teaser role below which I don't believe their version supports.

Naros:
Before I forget, @hashhar great work on this.  This was very much needed.

PR #288, opened 2019-05-02
DBZ-1234 DBZ-1234 Remove deprecated features and configuration options

Naros:
https://issues.jboss.org/browse/DBZ-1234

This requires PR https://github.com/debezium/debezium/pull/864 and should be included for the website when we publish version 0.10.


docs/connectors/mysql.asciidoc
```diff
@@ -1010,6 +1010,8 @@ Other than MySQL's `TIMESTAMP` data type, the MySQL temporal types depend on the
 [NOTE]
 ====
 As of Debezium 0.7 `adaptive_time_microseconds` mode was introduced and is the default `time.precision.mode` for the MySQL connector. Mode `adaptive` was marked as deprecated.
+
+As of Debezium 0.10 `adaptive` mode was removed and can longer be used.
```
gunnarmorling: So that's an interesting question: should we simply remove the option altogether? That'd be my preference altogether.

Going forward, we'll publish documentation specific to minor versions, making this problem go away. But as the options have been deprecated long time ago, they shouldn't be used for new set-ups anyways (also when using earlier connector versions still supporting them), and existing users shouldn't mind too much if we remove this from the docs.

PR #293, opened 2019-05-18
Add documentation to verify if the user have access to cdc table

yixin19:
The connector will not return warning message when the user doesn't have access to the cdc table, so it's better to verify it.

### DIFF
--- a/docs/connectors/sqlserver.asciidoc
+++ b/docs/connectors/sqlserver.asciidoc
@@ -76,6 +76,17 @@ EXEC sys.sp_cdc_enable_table
 GO
 ----
 
+Verify that the user have access to the cdc table
+[source, sql]
+----
+-- =========
+-- Verify the user of the connector have access, this query should not have empty result
+-- =========
+
+EXEC sys.sp_cdc_help_change_data_capture 
+GO
+----
+
 [[azure]]
 === SQL Server on Azure
 


docs/connectors/sqlserver.asciidoc
```diff
@@ -76,6 +76,17 @@ EXEC sys.sp_cdc_enable_table
 GO
 ----
 
+Verify that the user have access to the cdc table
+[source, sql]
+----
+-- =========
+-- Verify the user of the connector have access, this query should not have empty result
```
gunnarmorling: Nice improvement!

@jpechane, could we somehow check this programmatically during connector start-up?

jpechane: Already logged - https://issues.jboss.org/browse/DBZ-1281

jpechane:
@yixin19 Hi, do you have an explanation *why* the call returned empty list?

docs/connectors/sqlserver.asciidoc
gunnarmorling: Ok, sweet. Merging this one here.

gunnarmorling:
> Hi, do you have an explanation why the call returned empty list?

@jpechane, AFAIU, this returns information about capture instances and their config. If it returns an empty set, it indicates that there are no capture instances at all.

jpechane:
@gunnarmorling Well, the problem is that the capture instances were available and the capture tables were filled - just the method returned nothing - we spent few hours last week looking for the the problem. So I think it is useful to record why the information can be unavailable even if the CDC is correctly setup. My guess is security restirictions but it needs to be confirmed by @yixin19 who was discussing the matter with their db admin.

gunnarmorling:
Ok, gotcha. I'll add a pointer to that JIRA issue to capture the discussion here, while waiting to hear back from @yixin19. Thanks!

yixin19:
> @gunnarmorling Well, the problem is that the capture instances were available and the capture tables were filled - just the method returned nothing - we spent few hours last week looking for the the problem. So I think it is useful to record why the information can be unavailable even if the CDC is correctly setup. My guess is security restirictions but it needs to be confirmed by @yixin19 who was discussing the matter with their db admin.

To receive a result from the sp_cdc_help_change_data_capture, your user must have the SELECT permission on the table and must be assigned to the "cdc_role" database role.

It was a problem of permission for my case. Although the user has access to cdc table, il could happen that he doesn't have access to the capture instance. And sql server will not return any warning message as well. It is quite confusing for users.

Thanks again for your help @jpechane

PR #294, opened 2019-05-21
DBZ-1234 Doc update

gunnarmorling:
https://issues.jboss.org/browse/DBZ-1234

@Naros, here's the corresponding doc update to the option removal.


Naros:
LGTM

PR #298, opened 2019-05-24
Update documentation colun.mask.with.length.chars

knil-sama:
to explicit authorized value of length for column.mask.with.length.chars


gunnarmorling:
Interesting; why is it you're not simply black-listing such column instead of "masking" it with a zero-length string?

knil-sama:
Well, our use case required the structure to stay the same so we didn't wanted to drop columns and the 
column.truncate.to.length.chars really don't work with length = 0

gunnarmorling:
I see; thanks!

PR #299, opened 2019-05-27
DBZ-1235 Doc Updates

gunnarmorling:
@jpechane I've begun with the doc update around the source block changes. I've got two questions, I'll add inline comments for those. I'll push more commits tomorrow for the remaining connectors. Thanks!

### DIFF
--- a/docs/connectors/mongodb.asciidoc
+++ b/docs/connectors/mongodb.asciidoc
@@ -240,94 +240,111 @@ Let's look at what a _create_/_read_ event value might look like for our `custom
 {
     "schema": {
       "type": "struct",
-      "optional": false,
-      "name": "fulfillment.inventory.customers.Envelope",
-      "version": 1,
       "fields": [
         {
-          "field": "op",
           "type": "string",
-          "optional": false
+          "optional": true,
+          "name": "io.debezium.data.Json",
+          "version": 1,
+          "field": "after"
         },
         {
-          "field": "after",
           "type": "string",
-          "optional": true
+          "optional": true,
+          "name": "io.debezium.data.Json",
+          "version": 1,
+          "field": "patch"
         },
         {
-          "field": "patch",
-          "type": "string",
-          "optional": true
-        },
-        {
-          "field": "source",
           "type": "struct",
-          "name": "io.debezium.connector.mongo.Source",
-          "optional": false,
           "fields": [
             {
-              "field": "version",
               "type": "string",
-              "optional": false
-            },
-            {
-              "field": "name",
-              "type": "string",
-              "optional": false
-            },
-            {
-              "field": "rs",
-              "type": "string",
-              "optional": false
-            },
-            {
-              "field": "ns",
-              "type": "string",
-              "optional": false
-            },
-            {
-              "field": "sec",
-              "type": "int32",
               "optional": false,
+              "field": "version"
             },
             {
-              "field": "ord",
-              "type": "int32",
+              "type": "string",
               "optional": false,
+              "field": "connector"
             },
             {
-              "field": "h",
+              "type": "string",
+              "optional": false,
+              "field": "name"
+            },
+            {
               "type": "int64",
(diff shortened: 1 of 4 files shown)


docs/connectors/mongodb.asciidoc
```diff
@@ -240,94 +240,111 @@ Let's look at what a _create_/_read_ event value might look like for our `custom
...
+      ],
+      "optional": false,
+      "name": "dbserver1.inventory.customers.Envelope"
+      },
     "payload": {
-      "op": "r",
-      "ts_ms": 1465491411815,
-      "after": "{\"_id\":1004,\"first_name\":\"Anne\",\"last_name\":\"Kretchmar\",\"email\":\"annek@noanswer.org\"}",
+      "after": "{\"_id\" : {\"$numberLong\" : \"1004\"},\"first_name\" : \"Anne\",\"last_name\" : \"Kretchmar\",\"email\" : \"annek@noanswer.org\"}",
+      "patch": null,
```
gunnarmorling: Any idea why this shows up as `null` here now, @jpechane? I remember we discussed a similar question before, but what I don't remember is what the outcome of that was :)

docs/connectors/mongodb.asciidoc
```diff
@@ -240,94 +240,111 @@ Let's look at what a _create_/_read_ event value might look like for our `custom
...
       "source": {
         "version": "{debezium-version}",
+        "connector": "mongodb",
```
gunnarmorling: We had missed that before actually.

docs/connectors/mongodb.asciidoc
```diff
@@ -240,94 +240,111 @@ Let's look at what a _create_/_read_ event value might look like for our `custom
...
-          "type": "string",
-          "optional": false
-        },
-        {
-          "field": "after",
           "type": "string",
-          "optional": true
+          "optional": true,
+          "name": "io.debezium.data.Json",
+          "version": 1,
```
gunnarmorling: Any idea where this is coming from now, @jpechane? It wasn't there before.

jpechane: https://issues.jboss.org/browse/DBZ-100

docs/connectors/mongodb.asciidoc
jpechane: https://issues.jboss.org/browse/DBZ-984?focusedCommentId=13669063&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13669063

docs/connectors/mongodb.asciidoc
gunnarmorling: Sorry, I meant the "version" field, it wasn't there before. But then there was one for the Envelope type. This might have changed well before DBZ-1235 and never was updated in the docs. But I'm curious what's going on. The issue you linked seems not related?

jpechane: @gunnarmorling Looking for the vwriosn, I thought the type name was the problem so that's why I pointed to the origina l issue.

jpechane: @gunnarmorling Check Json datatype, it explicitly sets version to 1.

gunnarmorling: Ok, I see. Probably doesn't matter then. On the envelope version, this was consciously removed in DBZ-466 a while ago, we just never updated these examples here. All good.

docs/connectors/mongodb.asciidoc
gunnarmorling: Ah yes, thanks! I should bookmark that one ;) That said I'm wondering *why* the converter behaves that way; after all, there's a difference between not providing a value at all or explicitly setting it to null.

gunnarmorling:
Pushed one more commit for PG. @jpechane, good to go with that from my PoV.

jpechane:
@gunnarmorling +1

PR #300, opened 2019-05-27
DBZ-677 Updating docs for renamed SMTs

gunnarmorling:


jpechane:
@gunnarmorling Should not the left bar menu be changed too?

jpechane:
@gunnarmorling ping - regarding left menu bar

gunnarmorling:
Ah yes, good point. Perhaps you could do it tomorrow?

PR #301, opened 2019-05-28
DBZ-1224 Docs for SQL Server snapshot selects

gunnarmorling:


docs/connectors/sqlserver.asciidoc
```diff
@@ -625,6 +625,13 @@ Can be used to avoid snapshot interruptions when starting multiple connectors in
 |Specifies the maximum number of rows that should be read in one go from each table while taking a snapshot.
 The connector will read the table contents in multiple batches of this size. Defaults to 2000.
 
+|`snapshot.select.statement.overrides` +
+0.10.0 and later
```
jpechane: @gunnarmorling Should not the version string contain also also alpha/beta designation? We will release more 0.10.0 versions

jpechane:
@gunnarmorling Does not this work for the Oracle too?

gunnarmorling:
> Does not this work for the Oracle too?

It's not added yet to the config fields in `OracleConnectorConfig`. I'm never quite sure whether that'd still work then or not. In any case, it's not tested.

docs/connectors/sqlserver.asciidoc
gunnarmorling: @jpechane, no I don't think it's needed. People usually don't care about preview releases very much to begin with. And if they do, they'll take the latest one there is, so they won't have the issue of being on Alpha1 and trying to use an option only added in Beta1.

PR #311, opened 2019-06-10
Fix 0.10.0.Alpha2 release date

jpechane:


gunnarmorling:
Applied. No need for a PR really from my POV for this sort of change. Just apply these yourself, please :)

PR #317, opened 2019-06-17
Blog about protobuf decoder RPM

jpechane:


gunnarmorling:
Applied. Thanks a lot for the write-up and the RPM in the first place, @jpechane. Great work!

PR #320, opened 2019-06-30
DBZ-1284] Added docs for deployment with AMQ Streams

jcechace:
An initial draft of AMQ Streams installation docs.

* Links should be updated for AMQ Streams 1.2.0 
* I'm not sure where to put the information regarding PostgreSQL and Decoderbuf plugin. The plugin and its installation seem to be already well documented in other parts of the docs. Maybe mentioning that we only support Decoderbuf with AMQ Streams would be enough?


docs/amq-streams.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+
+Next we are going to verify whether the deployment is correct by emulating the link:/docs/tutorial/[Debezium Tutorial] and following the steps in link"/docs/openshift/#verifying_the_deployment"[OpenShift Installation]
+
+[NOTE]
+====
+You will want to change the pod names to correspond with your AMQ Streams deployment.
+====
+
+== Questions and Resources
+If you got any questions or requests related to running Debezium on OpenShift,
```
gunnarmorling: We should see what to link here. Either that new mailing list or some other channel. I'll clarify.

jcechace: Definitely, there is more place where links need to be added or updated.

gunnarmorling: Ok, so here is how the sentence should look like:

> In case of any requests or questions related to running the Debezium Developer Preview with AMQ Streams on OpenShift, please let us know by sending an e-mail to debezium-cdc-preview@redhat.com.

jcechace: Done

docs/amq-streams.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+We now have a working ```Kafka cluster``` running within ```OpenShift``` with ```Kafka Connect s2i```.
+
+After a while and depending on your Kafka configuration, these pods should be running.
+
+[source%nowrap,bash]
+----
+$ oc get pods
+
+NAME                                                      READY     STATUS      RESTARTS   AGE
+debezium-kafka-cluster-entity-operator-7b6b9d4c5f-k7b92   3/3       Running     0          5m
```
gunnarmorling: Where's the "debezium" name coming from? I think it'd help to add something like that to the sentence before:

> After a while and depending on your Kafka configuration, these pods should be running (assuming you used "debezium" as the AMQ Streams Kafka cluster name):

The contents in (...) is what I'm after, feel free to reword as you see fit.

jcechace: Done.

gunnarmorling:
Applied. Thanks a lot, @jcechace!

PR #327, opened 2019-07-09
Add documentation for PG custom snapshot modes

addisonj:
This adds the docs for the custom snapshot modes including details on
the SPI snapshotter interface.

### DIFF
--- a/docs/connectors/postgresql.asciidoc
+++ b/docs/connectors/postgresql.asciidoc
@@ -198,7 +198,7 @@ As mentioned link:#limitations[in the beginning], PostgreSQL 9.6 only supports l
 [[snapshots]]
 === Snapshots
 
-Most PostgreSQL servers are configured to not retain the complete history of the database in the WAL segments, so the PostgreSQL connector would be unable to see the entire history of the database by simply reading the WAL. So, by default the connector will upon first startup perform an initial _consistent snapshot_ of the database. Each snapshot consists of the following steps:
+Most PostgreSQL servers are configured to not retain the complete history of the database in the WAL segments, so the PostgreSQL connector would be unable to see the entire history of the database by simply reading the WAL. So, by default the connector will upon first startup perform an initial _consistent snapshot_ of the database. Each snapshot consists of the following steps (when using the builtin snapshot modes, *custom* snapshot modes may override this):
 
 1. Start a transaction with a https://www.postgresql.org/docs/9.6/static/sql-set-transaction.html[SERIALIZABLE, READ ONLY, DEFERRABLE] isolation level to ensure that all subsequent reads within this transaction are done against a single consistent version of the data. Any changes to the data due to subsequent `INSERT`, `UPDATE`, and `DELETE` operations by other clients will not be visible to this transaction.
 2. Obtain a `SHARE UPDATE EXCLUSIVE MODE` lock on each of the monitored tables to ensure that no structural changes can occur to any of the tables while the snapshot is taking place. Note that these locks do not prevent table `INSERTS`, `UPDATES` and `DELETES` from taking place during the operation.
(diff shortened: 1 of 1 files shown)


docs/connectors/postgresql.asciidoc
```diff
@@ -213,8 +213,96 @@ A second snapshot mode allows the connector to perform snapshots *always*. This
...
+The fourth snapshot mode, *initial only*, will perform a database snapshot and then stop before streaming any other changes. If the connector had started but did not complete a snapshot before stopping, the connector will restart the snapshot process and stop once the snapshot completes.
 
+The final snapshot mode, *custom*, allows the user to inject their own implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface via the `snapshot.custom.class` configuration property, with the class on the classpath of your Kafka Connect cluster (or included in the JAR if using the `EmbeddedEngine`). For more details, see the link:#custom-snapshot[Custom Snapshot] section.
+
+[[custom-snapshot]]
+=== Custom Snapshotter SPI
+For more advanced usages, the user can provide an implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface. This interfaces allows control of most of the aspects of how snapshots operate, such as whether to take a snapshot or not and the way the options used to open the snapshot transaction or take locks.
```
gunnarmorling:
```suggestion
For more advanced usages, the user can provide an implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface. This interface allows control of most of the aspects of how snapshots operate, such as whether to take a snapshot or not and the way the options used to open the snapshot transaction or take locks.
```

docs/connectors/postgresql.asciidoc
```diff
@@ -213,8 +213,96 @@ A second snapshot mode allows the connector to perform snapshots *always*. This
...
+        tableIds.forEach(tableId -> statements.append("LOCK TABLE ")
+                .append(tableId.toDoubleQuotedString())
+                .append(" IN SHARE UPDATE EXCLUSIVE MODE;")
+                .append(lineSeparator));
+        return Optional.of(statements.toString());
+    }
+}
+----
+
+All of the builtin snapshot modes are implemented in terms of this interface as well.
```
gunnarmorling: Can you add a sentence like:

> This interface is under active development and may change in future releases in potentially backwards-incompatible ways.

docs/connectors/postgresql.asciidoc
```diff
@@ -1423,7 +1511,11 @@ The following _advanced_ configuration properties have good defaults that will w
...
+|Specifies the criteria for running a snapshot upon startup of the connector. The default is *initial*, and specifies the connector can run a snapshot only when no offsets have been recorded for the logical server name. The *always* option specifies that the connector run a snapshot each time on startup. The *never* option specifies that the connect should never use snapshots and that upon first startup with a logical server name the connector should read from either from where it last left off (last LSN position) or start from the beginning from the point of the view of the logical replication slot. The *initial_only* option specifies that the connector should only take an initial snapshot and then stop, without processing any subsequent changes. Finally, if set to *custom* then the user must also set `snapshot.custom.class` which is a custom implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface. See link:#snapshots[snaphosts]
```
gunnarmorling:
```suggestion
|Specifies the criteria for running a snapshot upon startup of the connector. The default is *initial*, and specifies the connector can run a snapshot only when no offsets have been recorded for the logical server name. The *always* option specifies that the connector run a snapshot each time on startup. The *never* option specifies that the connect should never use snapshots and that upon first startup with a logical server name the connector should read from either from where it last left off (last LSN position) or start from the beginning from the point of the view of the logical replication slot. The *initial_only* option specifies that the connector should only take an initial snapshot and then stop, without processing any subsequent changes. Finally, if set to *custom* then the user must also set `snapshot.custom.class` option, referencing a custom implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface. See link:#snapshots[snapshots]
```

docs/connectors/postgresql.asciidoc
```diff
@@ -1423,7 +1511,11 @@ The following _advanced_ configuration properties have good defaults that will w
...
+
+|`snapshot.custom.class`
+|
+| A full java class name that must be an implementation of the `io.debezium.connector.postgresql.spi.Snapshotter` interface. Only used when `snapshot.mode` is *custom*
```
gunnarmorling:
```suggestion
| The fully-qualified name of a Java class implementing the `io.debezium.connector.postgresql.spi.Snapshotter` interface. Only used when `snapshot.mode` is *custom*
```

PR #332, opened 2019-07-25
fix disable cdc error

yixin19:


jpechane:
@yixin19 Aplied, thanks for spotting!

PR #333, opened 2019-07-25
[DBZ-393] Improve embedded documentation

willome:
- `server.id` replaced with `database.server.id` 
- minimal Executor instanciation


jpechane:
Applied, thanks @willome!

PR #334, opened 2019-07-25
DBZ-1035 Document new PostgreSQL Exported Snapshot feature

Naros:
https://issues.jboss.org/browse/DBZ-1035
This depends on https://github.com/debezium/debezium/pull/981


docs/connectors/postgresql.asciidoc
```diff
@@ -215,6 +215,8 @@ The third snapshot mode instructs the connector to *never* performs snapshots. W
 
 The fourth snapshot mode, *initial only*, will perform a database snapshot and then stop before streaming any other changes. If the connector had started but did not complete a snapshot before stopping, the connector will restart the snapshot process and stop once the snapshot completes.
 
+The fifth snapshot mode, *exported*, will perform a database snapshot based on the point in time when the replication slot was created.  This mode is useful when you want to perform snapshots in a lock-free way for a new connector.
```
gunnarmorling: This seems to downplay the advantages of this approach a bit. I mean, lock-free is great, no? Is there any negatives to using it (even as the default eventually), @Naros?

Naros: Good point; I didn't read the phrasing in that way.  As for negatives - I don't believe so.

PR #336, opened 2019-07-29
use correct casing: ZooKeeper instead of Zookeeper

erikdw:


Naros:
Hi @erikdw; thanks for the PR.  Merged.

PR #338, opened 2019-08-06
DBZ-766 Document new pgoutput decoder for PostgreSQL connector

Naros:


docs/connectors/postgresql.asciidoc
```diff
@@ -1314,6 +1341,13 @@ In such cases it is possible to switch to so-called *streaming* mode when every
 |`false`
 |Whether or not to drop the logical replication slot when the connector finishes orderly. Should only be set to `true` in testing or development environments. Dropping the slot allows WAL segments to be discarded by the database, so it may happen that after a restart the connector cannot resume from the WAL position where it left off before.
 
+|`publication.name`
+|`dbz_publication`
+|The name of the PostgreSQL publication created created for streaming changes when using `pgoutput`.
+
+This publication is created at start-up if it does not already exist to include _all tables_.  Debezium will then use its own white-/black- list filtering capabilities to limit change events to the specific tables of interest if configured.
```
gunnarmorling: Ah, that's interesting. I thought we'd rely this to be always created manually upfront?

gunnarmorling:
Rebased and applied. Thanks a lot, @Naros! I've added one more commit with a few more changes here and there. This also updates the _docs/install/postgres-plugins.asciidoc_ page (not sure whether you have seen that already, it might be of interest for your general site overhaul).

docs/connectors/postgresql.asciidoc
Naros: I believe we likely did make that final call and I had completely forgotten.  Do we want to make that adjust for the next 0.10 release or leave it as is @gunnarmorling ?

PR #339, opened 2019-08-09
DBZ-317 - Reorganized documentation by release.

Naros:
https://issues.jboss.org/browse/DBZ-317

I rebased this against the latest commits to develop as of 2019-08-09.  A few of the recent changes were not properly merged by git with files being moved around so they were added back as a part of commit https://github.com/debezium/debezium.github.io/commit/31b4e2590f95da0879773fa42ebb6ca0f52e8ddf


Naros:
@jcechace I added a version-selector option on the pages where it made sense; feel free to take a look and build the site and let me know if that works for you.  

@gunnarmorling I replaced the _Maven artifacts_ button per-release to be a drop-down selection where users could pick the specific connector plugin and it downloads it directly.  Let me know if that works for you.

Naros:
@gunnarmorling Added the canonical changes in https://github.com/debezium/debezium.github.io/pull/339/commits/0031ec2ba5ce465b9ccdfbd5dd5646a351e2efb6

gunnarmorling:
Awesome, thanks! Will review it tomorrow.

>

gunnarmorling:
Hey @Naros, it's looking great! 

> I replaced the Maven artifacts button per-release to be a drop-down selection where users could pick the specific connector plugin and it downloads it directly. Let me know if that works for you.

+1 That's very cool.

A few more comments:

* On the URLs: http://localhost:4242/docs/0.9/releases/, how about just having this under http://localhost:4242/docs/0.9/? After all, this page already *is* about one specific release, so having the "releases" in the URL seems not so sensible; WDYT?
* Instead of "Fifth bugfix of 0.9" etc., can we add the one or two top features of each release? There's a bit of the info for past releases in [JIRA](https://issues.jboss.org/projects/DBZ?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=released-unreleased); unfortunately, we got a bit sloppy on that end in 0.9. So it'd mean to go back to the release announcements (just for 0.9/0.10). Not for bugfix releases, there might nothing special to be said
* I think the breadcrumb should be *above* the specific page's title
* On http://localhost:4242/docs/0.9/releases/ (as said, I think better without the "releases") can we have the stable/development badge?
* A pity we can't have true 301 redirects for old URLs, but it seems that's how it is with GH Pages?
* On "How to get it", perhaps this section could just be replaced by another button in the releases table, e.g. "Maven Artifacts" and then that link to Maven Central?
* There seems to be no "stable" alias yet, e.g. /docs/stable/connectors/mongodb/ gives me a 404. I think we should have this one as the preferred target for external links (e.g. in the forum or on stack Overflow, and also the release specific documents should point to that as the canonical URL); the contents would be the same as the current stable (i.e. 0.9 right now)
* Can we have a link to the issue tracker on the 404 package, suggesting to file an issue about the broken URL (akin to how Hibernate is doing it)?
* It seems some (all?) canonical links are still missing, e.g. I don't see it in http://localhost:4242/docs/0.10/connectors/postgresql/; might be caused by the lack of that "stable" alias above

Please forgive me that wall of text, I hope it doesn't come across as too negative. It's great work overall, and not that much is missing from my PoV. Happy to discuss any details here or in the chat :) Thanks a ton!

gunnarmorling:
@Naros, also pushed one tiny commit to this branch with a few fixes.

gunnarmorling:
@Naros, one more general thought. What is concerning me a little bit is that the actual *documentation* still is maintained on the website. Instead, it might be preferable to have this as part of the code repository; that way, it'd be naturally versioned along with the actual code; also the PR flow would be more natural, as a code change PR wouldn't be merged unless it also updates the docs on newly added options etc.

Could you perhaps take an hour or so and see how [Antora](https://antora.org/) might fit into this (it's from Dan Allen and crew of AsciiDoc fame)? I think it handles versioned documentation very well, but of course it'd somehow have to be integrated with the other parts of the website. It's an open exploration, I'm not sure whether we should go for this or not, but I wanted to bring it up in any case.

Naros:
Hi @gunnarmorling, great catches and suggestions - I'll take care of those today.  About Antora; I do think that might be worth some exploration with a small PoC perhaps.  See my comments online about it https://gitter.im/debezium/dev?at=5d5423d77d56bc6080515afd

Naros:
> A pity we can't have true 301 redirects for old URLs, but it seems that's how it is with GH Pages?

I'll need to see if we can do it this way as I simply reused what we already had for this.

> There seems to be no "stable" alias yet, e.g. /docs/stable/connectors/mongodb/ gives me a 404. I think we should have this one as the preferred target for external links (e.g. in the forum or on stack Overflow, and also the release specific documents should point to that as the canonical URL); the contents would be the same as the current stable (i.e. 0.9 right now)

Ah that's slightly different than what I did but this should be doable.  In the PR as it stands right now, I simply added the old links like `/docs/connectors/mysql` to redirect to the latest stable link and then when we rendered `/docs/0.9/connectors/mysql` it would specify that its canonical url in the metadata was `/docs/connectors/mysql`.

If I understand correctly we then need to define 2 redirects:

    /docs/stable/connectors/mysql  ==>  /docs/0.9/connectors/mysql
    /docs/connectors/mysql  ==>  /docs/0.9/connectors/mysql

And then we need to make sure that the canonical URL in the `/docs/0.9/connectors/mysql` page actually refers to the `/docs/stable/connectors/mysql` link we're adding.  Let me know if I misinterpreted you.

> It seems some (all?) canonical links are still missing, e.g. I don't see it in http://localhost:4242/docs/0.10/connectors/postgresql/; might be caused by the lack of that "stable" alias above

Only the `0.9` pages have canonical URLs since the legacy links, e.g. `/docs/connectors/mysql` would point to whatever the latest stable series is, which is `0.9`.  The development version pages currently won't expose a canonical url, which I could see in hindsight as potentially problematic?

Should we have a `/docs/development` or `/docs/unstable` canonical URL that the `0.10` series pages use?  And when we introduce `0.11` and `0.10` goes stable, what canonical URL, if any, should the `0.9` pages specify?

gunnarmorling:
@Naros, yes, I think a quick PoC of Antora for the docs would be useful. Let's shelf the canonical discussion until after that I'd say.

.travis.yml
```diff
@@ -1,10 +1,15 @@
 language: ruby
```
jpechane: @Naros I think the language should be switched to https://docs.travis-ci.com/user/languages/minimal-and-generic/ as the build is delegated to the Docker

.travis.yml
```diff
@@ -1,10 +1,15 @@
...
+  - 2.3
+services:
+  - docker
 branches:
   only:
   - develop
-script: rake travis
+before_install:
+  - docker pull debezium/awestruct
+  - docker run --rm --user $UID:$GID -e LC_ALL=C.UTF-8 -e LANG=C.UTF-8 -v $(pwd):/site debezium/awestruct rake setup && rake travis
```
jpechane: @Naros Not sure about this one but as it is written I expect it will execute `rake setup` in context of Docker container and then `rake travis` as a process outside of container

gunnarmorling:
Hey @Naros, just built myself locally to do some testing. It looks great overall! A few last remarks:

- [ ] /docs/releases-overview/
  - [ ] Cassandra missing under "Tested With"
  - [ ] "Tested With" should be "Tested Versions" and there should be a sentence like "The Debezium connectors have been tested with the following versions of Java, Apache Kafka (Connect) and the databases. Note that other database versions not listed here may be supported, too.
  - [ ] MySQL 5.7 should be added
- [ ] Different places: 0.10 comes before/above 0.9; can we do it the other way around?
- [ ] docs/0.10/ (same for 0.9)
  - [ ] There should be some space between the caption and the badge
  - [ ] The HTML page title should be something like "Debezium Release Series 0.10" instead of "Index"
- [ ] There still is some general confusion around the URLs for docs vs. releases vs. documentation; here's what I suggest (in nutshell, only have documentation and releases):
  - [ ] /docs/releases-overview/ -> /releases
  - [ ] /docs/0.10/ -> /releases/0.10; same for 0.9
  - [ ] /docs/contribute -> /community/contribute
  - [ ] /docs/code-of-conduct -> /community/code-of-conduct
  - [ ] /docs/roadmap -> /roadmap
  - [ ] /docs/online-resources -> /documentation/online-resources
  - [ ] /docs/faq -> /documentation/faq
  - [ ] Instead of /documentation/debezium/0.10/, can we do just /documentation/0.10/; i.e. we'd have the common resources and the version-specific docs at the same level: /documentation/{faq|online-resources|0.9|0.10}; WDYT?
  - [ ] /docs -> /documentation; should have links to the different stable and release-specific assets (the latter driven by the YAML with the stable/dev badges perhaps?)
- [ ] /community/: missing the menu on the left; also let's remove the reference to 2016
- [ ] Footer: broken links to architecture and feature pages; other links need updating as per above

On the Antora pages themselves:

- [ ] /documentation/debezium/0.10/
  - [ ] Cassandra missing under connectors
  - [ ] Remove "Welcome" menu entry, it's the same as when clicking on "Debezium Documentation"
  - [ ] "Welcome to the Debezium documentation!" -> "Debezium Documentation"

gunnarmorling:
Also pushed one commit with a small fix.

Naros:
@gunnarmorling I believe I got all the changes either pushed to this PR or the relevant documentation PRs in the main repo to cover your recommendations.  Now to finalize the travis integration so that it works as desired and I believe this should be all set.

gunnarmorling:
@Naros, fantastic! Merged the main PRs and pushed a commit here for the changed 0.9 branch name. It's almost perfect, here's only one more thing I just noticed: 

- [ ] /releases/  highlights the "Documention" button in the top menu; I had wondering that before: should we have "Releases" in the top menu, too, in fact I think that'd be warranted.

And this one still seems missing:

- [ ]   /documentation; should have links to the different stable pages and release-specific reference docs (the latter driven by the YAML with the stable/dev badges perhaps?)

If you prefer, we can go live with out the latter, though.

We're getting there!

PR #341, opened 2019-08-15
Postgres connector docs fix on TIMESTAMP mapping

gunnarmorling:
@jpechane, minor doc fix which I noticed while PR reviewing.


jpechane:
@gunnarmorling Applied, thanks!

PR #347, opened 2019-08-31
Update to Strimzi 0.13.0

scholzj:
The OpenShift tutorial is currently using Strimzi 0.9.0. This PR updates it to the latest version - 0.13.0. It also changes how the Kafka and Kafka Connect are deployed from the templates (the `oc new-app` does not work on OCP4 because it deploys a custom resource) - the process works also with 3.11. 

Additionally it removed the mentiones of OpenShift Online and Dedicated where Strimzi doesn't work for quite a long time due to the limitations of these environments (not possible to install CRDs and Cluster Roles etc.)

### DIFF
--- a/_config/site.yml
+++ b/_config/site.yml
@@ -26,7 +26,7 @@ asciidoctor:
     debezium-docker-label: 0.9
     confluent-platform-version: 5.1.2
     debezium-kafka-version: 2.3.0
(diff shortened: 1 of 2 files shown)


gunnarmorling:
LGTM; thanks, @scholzj!

PR #350, opened 2019-09-04
DBZ-1458 Adding architecture page

gunnarmorling:
@jpechane, @Naros, as dicussed last week, here's the architecture page for the website.

https://issues.jboss.org/browse/DBZ-1458


gunnarmorling:
Pushed something for the features page, too. Let me know if there's something you think should be added.

gunnarmorling:
About the "concept" page, I'm not sure; perhaps it should just be removed? Wouldn't know what to add there, which isn't yet on "architecture" and "features". WDYT?

Naros:
@gunnarmorling I completely agree with removing _concepts_.  I think having the features and architecture pages should suffice for what we need as we can always add to it later on.

gunnarmorling:
Ok; could you then perhaps add a commit for removing the page + menu item
and push it to this PR? Thanks!

>

gunnarmorling:
@Naros, I reckon we can close this as it's incorporated in your docs rework PR #339?

Naros:
That's correct, closing this PR.

PR #357, opened 2019-09-25
Added support for right navbar for TOC on large screens

Naros:
This renders documentation pages with a TOC section in a right navigational bar that is responsive based on the screen size.  On tablet and mobile devices, the TOC continues to be rendered at the top of the asciidoc page like it always has.

### DIFF
--- a/_antora/supplemental_ui/css/debezium-antora.css
+++ b/_antora/supplemental_ui/css/debezium-antora.css
@@ -109,12 +109,12 @@ table.tableblock {
 
 /* Render article with full width */
 .doc {
-  max-width: 95%;
+  max-width: 70%;
 }
 
 /* Keep images scaled down a bit */
 .doc .image img {
-  max-width: 600px;
+  max-width: 100%;
 }
 
 @media screen and (min-width: 1200px) {
@@ -140,12 +140,26 @@ table.tableblock {
     top: 148px;
     right: 0;
     color: #222;
+    overflow-y: auto;
+    height: calc(100vh - 8.25rem);
+  }
+
+  .doc.with-toc .toc::-webkit-scrollbar {
+    width: .25rem;
+  }
+
+  .doc.with-toc .toc::-webkit-scrollbar-thumb {
+    background-color: #c1c1c1;
   }
 
   /* Keep title hidden */
   .doc.with-toc .toc > .title {
     display: none;
     height: 0;
+  }
+
+  .doc.with-toc .toc ul {
+    padding-left: 20px;
   }
 
   .doc.with-toc .toc > ul {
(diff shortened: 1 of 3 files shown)


gunnarmorling:
@Naros, very nice, I like it overall. A few comments:

* Can pages without a ToC span the full width? It looks a bit odd to have all the large white space on the right in this case.
* Can the ToC be moved further to the right? There's quite a fair bit of white space next to it on the right side
* Instead of having a full box around the ToC, how would it look like with the separator line just on the left edge of that box, akin to the Bootstrap example you shared. Perhaps give this a try and share a screenshot of how that would look like?

Thanks again!

Naros:
@gunnarmorling So I reworked all the CSS changes I made and I think this actually comes out being a much better implementation all around on modern browsers.  I was able to incorporate all the requested changes.

gunnarmorling:
@Naros, definitely looking better. I still have a few issues, though:

* The image on /documentation/reference/0.10/architecture.html is very large now (compare to prod)
* I feel the right menu still has too much whitespace on the right side and I think it takes up too much space in general; apart from the whitespace, perhaps the fontsize should also be a bit smaller? I can't find that bootstrap example you shared, but I think the looks of the left and right menus were a bit different from each other, which I found helped to make clear that one is the overall navigation and the other one is the intra-page navigation

Screen Shot 2019-09-27 at 21 25 26

* I cannot access any ToC entries if they are out of the view port; I guess that's just the way it is given it's sticky, but then I also cannot collapse the entries, unlike implied by the arrows used in the menu. Would be interesting to see how this is handled in that example you shared?

Naros:
@gunnarmorling What browser are you testing with?  Chrome, Firefox, other?

gunnarmorling:
This is with Firefox 69.0.1.

>

Naros:
Alright I'll keep checking but that isn't what I observe in both Chrome and Firefox.  In fact, I see in prod that the image is 1260x700 while in the PR its slightly smaller, 1170x557.  I am specifying the size in rem units and perhaps that's why.  You don't happen to have the browser zoom set to something besides 100% do you?

gunnarmorling:
Hum, no, I don't think I have set it to anything other than 100%.

>

Naros:
No worries, feel free to give the latest push a glance as time permits.

gunnarmorling:
@Naros, yeah, looking *much* better now! Some last nits:

* The carets used as icons in the right-hand side menu suggest it would be collapsable, like the left menu, but it isn't. Can we find another icon not implying that?
* There's no scrollbar for the right menu, also if it goes beyond the view port (compare to the left side menu). No critical problem I think right now, as there are not that many entries in any ToC (yet); so could be done later on, but it might become an issue eventually
* The architecture picture is too *small* now (go figure ;) See below:

Screen Shot 2019-09-30 at 20 51 59

Naros:
> The carets used as icons in the right-hand side menu suggest it would be collapsable, like the left menu, but it isn't. Can we find another icon not implying that?

Do any of the following ideas work for you?

Using dashes:
Option-Dash

Using squares:
Option-block

No image:
Option-noimage

> The architecture picture is too *small* now (go figure ;)

Would doubling its width then be too big?  If so, we scale it up by another 50%, from 600px to 900px.
Do you think that would work?

jpechane:
@Naros My two cents - the dashes in the first image art too intrusive for me. Second and third are both good from my POV

Naros:
I believe I'm going to go with the no image solution because I think if we find oureselves with a grand-child layer or great-grand-child layer in the TOC, it just looks like it flows better rather than having the top point in each section with an image.

gunnarmorling:
> I believe I'm going to go with the no image solution

👍 

On the architecture image, dunno. I really like how it's currently done on the live website. In particular, the text lines don't flow super-wide either, but there's a maximum width in place. Which i think makes sense, reading such long lines gets very hard. Can't this just stay as it is?

Naros:
It probably can; let me see what CSS magic we can do.  So essentially we don't want to change the max-width used by the text/images in the asciidocs, we just want the right navigation bar flush on the right side of the browser window?

gunnarmorling:
@Naros, I'm a bit concerned we're getting lost in coming up with our own solution for something which should already have been resolved elsewhere. You once shared the Bootstrap docs, and I think they did it really well (e.g. their menu isn't attached to the right of the viewport, but sticks to the text if the viewport is too wide. That makes lots of sense to me in terms of usability. So can we do that?

Naros:
Oh I completely misunderstood what you wanted.  So just to be clear, is this what you're after?

SQLServer

gunnarmorling:
Yes, hat looks great!

>

PR #363, opened 2019-10-04
Adding a link to my recent Kafka Summit presentation slides

jbfletch:
I used Debezium in my example scenario in my recent Kafka Summit presentation. @gunnarmorling asked if I would open a PR to add a link to my slides.


gunnarmorling:
Awesome; thanks a lot!

PR #364, opened 2019-10-07
DBZ-1526 Add google analytics to Antora generated pages

Naros:


gunnarmorling:
Applied. Thanks, @Naros!

gunnarmorling:
@Naros, something is odd. See attached image, do you observe this, too:
Screen Shot 2019-10-09 at 16 49 16

PR #366, opened 2019-10-12
Update users.asciidoc

ayush-chauhan:


gunnarmorling:
Thanks a lot four adding OYO! Could you only rework the commit so it only lists the company name? We've considered linking, but then it'd seem odd to do it for some but not for others. That is, we could add a "details" link as for some of the other cases, where there's a blog with details on the specific Debezium usage. Thanks again!

gunnarmorling:
Applied. Thanks a lot, @ayush-chauhan!

PR #369, opened 2019-10-16
DBZ-892 Document postgres large WALs

jpechane:
https://issues.jboss.org/browse/DBZ-892


documentation/faq.asciidoc
```diff
@@ -275,3 +275,27 @@ Caused by: java.lang.IllegalStateException: Cannot read the binlog filename and
...
+
+== Why Debezium PostgreSQL connector causes abnormal consumption of WAL database disk space?
+
+Two reasons that explain the situation
+
+ * Debezium regularly confirms LSN of processed events to the database.
+This is visible as `confirmed_flush_lsn` in `pg_replication_slots` slots table.
+The database is responsible for reclaiming the disk space and the WAL size can be calculated from `restart_lsn` of the same table.
+So if the `confirmed_flush_lsn` is regularly increasing and `restart_lsn` lags then the database does need necessity to reclaim the space.
+Disk space is usually reclaimed in batch blocks.
```
gunnarmorling: I don't quite understand this first item. Will this be resolved automatically when PG reclaims the next batch? I'd expect this not to cause growth in the Gigabyte of ranges, or is it like that? A sentence should be added that this is expected then and no action is needed as PG will handle this periodically by itself.

documentation/faq.asciidoc
```diff
@@ -275,3 +275,27 @@ Caused by: java.lang.IllegalStateException: Cannot read the binlog filename and
...
+ * The PostgreSQL instance contains multiple databases where one of them is a high-traffic database.
```
gunnarmorling: There's a third case missing which should be added: traffic in the same DB as the slot is registered for, but only on non-whitelisted schemas/tables.

gunnarmorling:
Two comments inline. I'd suggest this to go into the connector docs actually, as it's an important matter to consider when setting up the connector. And then have a reference from the FAQ to there, i.e. leave the question, but the answer is the reference to the docs.

Really, really happy to see this finally being sorted out, thanks a lot!

documentation/faq.asciidoc
jpechane: @gunnarmorling No it should not. If it is the case then it is a bug in Debezium. The non-whitelisted events still should trigger LSN flush via heartbeats.

documentation/faq.asciidoc
jpechane: @gunnarmorling The problem is that we've seen reports that the wal grows till GBs before it is reclaimed. I'll add the sentence.

documentation/faq.asciidoc
jpechane: @gunnarmorling OTH I'll add it as it requires heartbeats to be enabled.

gunnarmorling: Sorry, I should have clarified: this situation is a problem if heartbeats are disabled. Which is likely, as that's the default.

jpechane:
Moved to main docs - https://github.com/debezium/debezium/pull/1082

gunnarmorling:
Applied; thanks, @jpechane!

PR #372, opened 2019-10-20
added blog about audit log with kogito admin service

mswiderski:


gunnarmorling:
Really great stuff, thank you very much, @mswiderski! I've pushed one more commit with some last wording adjustments. Merging it now, will be up in a bit. Thanks again!

PR #377, opened 2019-11-11
Trivial grammar fix.

markdoliner:
"their" ==>"they're"


gunnarmorling:
Agh, good catch. Thx!

PR #378, opened 2019-11-11
DBZ-1466 Add author mode support for locally building

Naros:
https://issues.jboss.org/browse/DBZ-1466

In order to use author mode, two things will be necessary:

1. Start docker mapping a local checked out copy of the Debezium main code repository as a volume using the name `/debezium`, e.g. `-v <local_path>:/debezium`
2. When running rake, also provide it with the `author` task, e.g. `rake clean author preview`.


jpechane:
@Naros Nice feature, applied!

PR #379, opened 2019-11-13
Pipedrive is using debezium

pimpelsang:


jpechane:
@pimpelsang Thanks for adding!

PR #383, opened 2019-11-22
Fix Segment/Ctlstore link

collinvandyck:
Markdown is easy!


gunnarmorling:
Ha, yeah, also constantly tripping over different link styles in Markdown vs. AsciiDoc vs. JIRA :)

PR #386, opened 2019-12-02
DBZ-1640 Provide Antora UI locally

Naros:
https://issues.jboss.org/browse/DBZ-1640


gunnarmorling:
+1 for becoming independent of that resource being available. That said, we also cut off ourselves from updates done upstream. Not sure, do they happen often? Is it a concern, or something which we just can do every now and then? WDYT?

Naros:
@gunnarmorling I'm not sure how often they officially cut releases either but looking at the commit history, I think its reasonable for us to manually update it once a month or so.

PR #394, opened 2019-12-12
Blogpost about externalized secrets

jpechane:


gunnarmorling:
Excellent! Added one commit to sprinkle a few "the"s and the like and merged. Thanks a lot, @jpechane!

PR #401, opened 2020-01-06
Dbz 1668

Naros:
https://issues.redhat.com/browse/DBZ-1668

### DIFF
--- a/_antora/supplemental_ui/css/debezium-antora.css
+++ b/_antora/supplemental_ui/css/debezium-antora.css
@@ -324,7 +324,7 @@ table.tableblock {
 }
 
 @media screen and (min-width: 2560px) {
-  .doc.with-toc {
+  .doc.with-toc, .doc {
     max-width: 100rem;
   }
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hey, this doesn't seem to change much for me. My screen width is 2880 FWIW; it still looks exactly like in the screenshot I shared in the JIRA issue.

Naros:
Ah; then your monitor width is in between the standard break points I added, so I'll need to add more breakpoints I suppose.  I just added 2560/3840 but looking at https://en.wikipedia.org/wiki/Graphics_display_resolution, I probably should add a 3200 and 3440 breakpoint.  I'll add those and let you check.

gunnarmorling:
So I'm wondering whether those fixed break points are the best approach. Perhaps the right-hand sight menu should just be attached to the right side of the window, giving all scaling wider and narrower space to the middle part with the actual contents. E.g. as done in the Bootstrap docs: https://getbootstrap.com/docs/4.1/getting-started/introduction/. WDYT?

Naros:
My memory is a bit foggy but I thought we went down that path sometime back in August or so when we first attempted the ToC changes and there was a reason we went with this approach.  I'd have to look back through the PR comments to see, but I can certainly look into seeing if we could make that work instead.

gunnarmorling:
Yes, that would be my preference. Unfortunately, I also don't quite remember why we didn't go for this back then. But it seems just so wasteful if that space on the right isn't used why the main column is so narrow.

Naros:
This PR is superseded by https://github.com/debezium/debezium.github.io/pull/409, closing in favor of the new implementation in the new PR.

PR #402, opened 2020-01-13
Add Synaltic to users page

yabir:


gunnarmorling:
Very nice, thanks a lot! Could you only rework the commit so to keep the list in alphabetic order and also author the commit so it uses an e-mail associated to your GitHub account? Thanks again!

yabir:
Thanks for the quick reply @gunnarmorling ! I dropped the first commits and Added another one with the modifications you suggested + Synaltic's blog post

gunnarmorling:
Applied, thx a lot. Will be live in a few min.

PR #409, opened 2020-01-17
DBZ-1668 Documentation rendering enhancements

Naros:
https://issues.redhat.com/browse/DBZ-1668

This PR has a sibling PR in the main repo, https://github.com/debezium/debezium/pull/1216.  We should make sure that the main repo PR is merged prior to this.  

* Reworked Antora UI
  * ToC is now flush to right of window
  * Documentation content uses responsive layout to maximize render space
  * Removed all obsolete responsive CSS for wider screen widths
* Asciidoc changes
  * Tables are now stripped and rendered with grid orders
  * Table headers now have a darker background color
  * Tables rendered using horizontal scrollbar when overflowing
  * Rendering of other blocks such as listings and examples to be consistent
* HighlightJS changes
  * Improved color rendering for Java and and XML

This PR supersedes https://github.com/debezium/debezium.github.io/pull/401.

### DIFF
--- a/_antora/supplemental_ui/css/debezium-antora.css
+++ b/_antora/supplemental_ui/css/debezium-antora.css
@@ -29,8 +29,71 @@ table.tableblock {
   border-radius: 3px;
 }
 
+table.tableblock, th.tableblock, td.tableblock {
+  border: 0 solid #e1e1e1;
+}
+
 .tableblock th.tableblock {
   font-weight: bold;
+}
+
+table.frame-all {
+  border-width: 1px;
+}
+
+table.grid-all th.tableblock, table.grid-all td.tableblock {
+  border-width: 0 1px 1px 0;
+}
+
+table.tableblock thead,
+table.tableblock tfoot {
+  background: #f0f0f0;
+  font-weight: bold;
+}
+
+.doc .table-scroll-wrapper {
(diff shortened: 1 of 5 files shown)


gunnarmorling:
@Naros, nice, tables look much better now. It seems I'm missing one thing, though:

> ToC is now flush to right of window

That's not the case for me, it's fixed ToC at the top of the page:

Screen Shot 2020-01-28 at 09 59 06

Tbh. that's fine for me if we can't get the right-hand side option under control, but as per your comment above I'm wondering whether something is missing from this PR, or I'm doing something wrong. Thanks!

Naros:
> That's not the case for me, it's fixed ToC at the top of the page:

It should only be fixed at the top of the page with 1200 width or below, which I assume isn't the case?

gunnarmorling:
> which I assume isn't the case?

No, it's not (see the width of the screenshot I shared, it's > 2500
pixels). But I have a high-res (Retina) display, so perhaps that messes
with the number processed by that CSS selector? I really have no idea about
all that stuff ;)

>

Naros:
@gunnarmorling Could you try rebuilding this again and force-refresh the page?  I just double checked this locally and I can't reproduce it so I'm wondering if this might be a caching issue or possibly a javascript error with your browser?

PR #413, opened 2020-02-04
1.1 prep

Naros:
This PR goes in tandem with https://github.com/debezium/debezium/pull/1237.  This adds the needed code to support the DB2 connector from the releases and the tested matrix.  

@jpechane I also added a shell of the `1.1.0.Beta1.yml` which you can update with pertinent 1.1 release details as needed when releasing 1.1.

### DIFF
--- a/_config/site.yml
+++ b/_config/site.yml
@@ -109,6 +109,8 @@ profiles:
         name: SQL Server
       cassandra:
         name: Cassandra
+      db2:
+        name: Db2
   staging:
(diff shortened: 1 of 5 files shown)


gunnarmorling:
LGTM; can we merge this already?

Naros:
@gunnarmorling We could but it would show the `1.1.0.Beta1` on the releases page by doing so and the download links would point to artifacts which are not yet available.  Up to you though.

gunnarmorling:
Gotcha, makes sense. Thx!

PR #414, opened 2020-02-05
Adding Behalf to Debezium in production use list

lirank:


gunnarmorling:
Thanks for the addition! Nice write-up, too.

lirank:
Thank you Gunnar!

PR #417, opened 2020-02-07
Removing inline styles from images

gunnarmorling:
Hey @Naros, can you check out and merge this one if you don't see any issues? Intention is to get rid the inline CSS (margin etc.) around images. I checked a few pages and it looked good.

All the styling is in CSS now, so that's better. To give some context: the raw HTML is done to assign the "responsive-image" class to images. Ideally that all shouldn't be needed to begin with as AsciiDoctor should make images responsive itself, but for whatever reason it doesn't in our case. Unfortunately, there's also no way to give a class for an AsciiDoctor `image`, hence the fall-back to raw HTML. But I think it's acceptable with this fix here.

Note this should be preferred:

```
[.imageblock.centered-image]
====
++++
<img ...>
++++
====
```

Over 

```
++++
<div class="imageblock centered-image">
    <img ...>
</div>
++++
```

So to minimize usage of raw HTML. The latter form is still used in quite a few places, no need to update, but the former should be used going forward.

### DIFF
--- a/blog/2018-08-30-streaming-mysql-data-changes-into-kinesis.adoc
+++ b/blog/2018-08-30-streaming-mysql-data-changes-into-kinesis.adoc
@@ -24,10 +24,12 @@ which in our case will convert the record into its JSON representation and submi
 
 The overall architecture looks like so:
 
-[.centered-image.responsive-image]
+[.centered-image]
+====
 ++++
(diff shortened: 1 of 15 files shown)


gunnarmorling:
@Naros, any thoughts?

Naros:
@gunnarmorling LGTM, merging.  Thanks!

PR #422, opened 2020-02-15
Fix repo url in README

hashhar:
The repository URL seems to be leftover from an older repo.


hashhar:
Never mind, somehow GitHub magically makes the older URL also work. :tophat:

gunnarmorling:
> GitHub magically makes the older URL also work

It's still a good change. Can you re-open it? We should give the correct, current name.

gunnarmorling:
Thx!

PR #426, opened 2020-02-20
DBZ-1694 Document column swap

jpechane:
https://issues.redhat.com/browse/DBZ-1694

### DIFF
--- a/documentation/faq.asciidoc
+++ b/documentation/faq.asciidoc
@@ -280,12 +280,12 @@ The solution is to indirectly enable the `log_bin` property, activating certain 
 == Why Debezium PostgreSQL connector causes abnormal consumption of WAL database disk space?
 See link:/documentation/reference/1.0/connectors/postgresql.html#wal-disk-space[WAL Disk Space Consumption] in PostgreSQL connector documentation.
 
-== Why a new schema version is not created when the content of two columns in a table is swapped?
-If two columns in a table are swapped in that way that after the change the table schema is same as before then no new version of schema is not created in schema registry.
+== Why is no new schema version created when the content of two columns in a table is swapped?
+If two columns in a table are swapped in that way that after the change the table schema is same as before, then no new version of the schema gets created in the schema registry.
 An example of such operation could be:
 
  * original table - `id`, `c1`, `c2` where `c1` and `c2` are of the same type
  * column swap - `id`, `c2`, `c1`
  * column rename - `id`, `c1`, `c2`
 
-The schema registry creates a new version of schema only if the schema logically changes but in this case the schema is the same after the change for an external observer.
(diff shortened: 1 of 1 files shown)


documentation/faq.asciidoc
```diff
@@ -279,3 +279,13 @@ The solution is to indirectly enable the `log_bin` property, activating certain
 
 == Why Debezium PostgreSQL connector causes abnormal consumption of WAL database disk space?
 See link:/documentation/reference/1.0/connectors/postgresql.html#wal-disk-space[WAL Disk Space Consumption] in PostgreSQL connector documentation.
+
+== Why a new schema version is not created when the content of two columns in a table is swapped?
```
gunnarmorling:
```suggestion
== Why is no new schema version created when the content of two columns in a table is swapped?
```

documentation/faq.asciidoc
```diff
@@ -279,3 +279,13 @@ The solution is to indirectly enable the `log_bin` property, activating certain
...
+If two columns in a table are swapped in that way that after the change the table schema is same as before then no new version of schema is not created in schema registry.
+An example of such operation could be:
+
+ * original table - `id`, `c1`, `c2` where `c1` and `c2` are of the same type
+ * column swap - `id`, `c2`, `c1`
+ * column rename - `id`, `c1`, `c2`
+
+The schema registry creates a new version of schema only if the schema logically changes but in this case the schema is the same after the change for an external observer.
```
gunnarmorling:
```suggestion
The schema registry creates a new version of schema only if the schema logically changes, but in this case the schema is the same after the change for an external observer.
```

documentation/faq.asciidoc
```diff
@@ -279,3 +279,13 @@ The solution is to indirectly enable the `log_bin` property, activating certain
 
 == Why Debezium PostgreSQL connector causes abnormal consumption of WAL database disk space?
 See link:/documentation/reference/1.0/connectors/postgresql.html#wal-disk-space[WAL Disk Space Consumption] in PostgreSQL connector documentation.
+
+== Why a new schema version is not created when the content of two columns in a table is swapped?
+If two columns in a table are swapped in that way that after the change the table schema is same as before then no new version of schema is not created in schema registry.
```
gunnarmorling:
```suggestion
If two columns in a table are swapped in that way that after the change the table schema is same as before, then no new version of the schema gets created in the schema registry.
```

Does this actually only apply to the schema registry, or also to the schema embedded in JSON events?

jpechane: @gunnarmorling JSON schemas are not versioned, they are not kept anywhere so there is no way how to version them

jpechane:
@gunnarmorling I've applied your suggestions

PR #431, opened 2020-03-02
Fix typos in "pgoutput"

laurenty:


gunnarmorling:
Good catch! Weird, the same odd mistake twice :) Thanks!

PR #435, opened 2020-03-11
DBZ-1798 Fix miscellaneous documentation issues

Naros:
https://issues.redhat.com/browse/DBZ-1798

In commit 27645e8, I am forcing the Travis CI build step to generate a `.nojekyll` file in the pre-build steps since Awestruct does not copy the `.nojekyll` file from the root of the repository.  This should allow images stored in `/documentation/reference/_images` to be rendered as they'll be published to github pages where-as previously without the `.nojekyll` they were not.

In commit 22248d2, this should resolve the overflow issue when browser window size is smaller than 1600px, forcing the image to respect aspect ratio and resize appropriately based on resolution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_install:
   - rm -rf ./_antora_cache
   - docker ps -a
   - ls -la ./_site/documentation
+  - touch ./_site/.nojekyll
 script: rake travis
 env:
   global:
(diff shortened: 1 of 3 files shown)


gunnarmorling:
>  I am forcing the Travis CI build step to generate a .nojekyll

Nice trick :)

gunnarmorling:
Let's see whether it helps; to be verified in a few min.

PR #461, opened 2020-06-01
DBZ-1951 Introduce banner for development/older versions

Naros:
https://issues.redhat.com/browse/DBZ-1951

So I decided to base the logic on when to apply the version banner because I wanted to see if we can do this without introducing even more steps to the release process for maintaining attributes. Antora maintains an array of versions and we use this array to compare against the version of the page rendered to see whether to apply the "development" or "outdated" banners.  

The banner placement is probably not ideal at the moment since it scrolls with the page content but in order to avoid problems with other rendering concerns, we can't make it static above the breadcrumbs portion of the page since the content that is scrollable is offset by a fixed number of pixels.

I'm open to any feedback or suggestions.

### DIFF
--- a/_antora/supplemental_ui/css/debezium-antora.css
+++ b/_antora/supplemental_ui/css/debezium-antora.css
@@ -304,6 +304,26 @@ strong {
   color: #e6e87c;
 }
 
+div.version-banner {
+    padding: 0.75rem 1.25rem;
+    margin-bottoM: 1rem;
+    border: 1px solid transparent;
+    border-radius: 0.25rem;
+    margin-top: 25px;
+}
+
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Looks great. Re placement, I think that's just fine. In fact, I actually dislike the sticky banners elsewhere, as I don't need to have this hint jump into my face all the time, reading it once is good enough :) The one downside is that one might miss it when arriving to a specific sub-section via an anchor link, but IMO that's acceptable.

What's really cool is that the "go here" link is specific to the actual page. I first thought it'd yield in a broken link on a new page added *only* in dev docs, but there it points to the stable doc's root. Very nice!

PR #475, opened 2020-07-14
Added Heroku's streaming data connectors

bibryam:


gunnarmorling:
Applied; Thx, @bibryam!

PR #492, opened 2020-09-15
DBZ-78 revert JSON fix: numbers get passed as numbers

rk3rn3r:


gunnarmorling:
Why not; but does it make any practical difference?

rk3rn3r:
Else we need to fix all the config in the texts as they are without. You also forgot fixing the last JSON. Feels a bit like panicing when this was changed.

PR #500, opened 2020-10-01
Add GoHealth to Debezium users

petoju:
Adding our company.


jpechane:
@petoju Thanks a lot!

PR #507, opened 2020-10-22
Adding 1.4 Alpha1 announcement

Naros:


gunnarmorling:
Nice, thx! I'll keep that for tomorrow as discussed, so to not overwhelm
users with too many news on a single day.

>

blog/2020-10-22-debezium-1-4-alpha1-released.adoc
```diff
@@ -0,0 +1,96 @@
...
+== Vitess Connector
+
+https://www.vitess.io[Vitess] is a database solution for deploying, scaling, and managing large clusters of MySQL.
+We are excited that the Vitess engineers decided to build a CDC solution based on Debezium and to https://www.github.com/debezium/debezium-connector-vitess[open-source it] under the Debezium umbrella.
+This connector is released in *incubating* state in Debezium 1.4.
+
+A blog post with more details around this connector will follow every soon;
+in the mean time please refer to the connector https://debezium.io/documentation/reference/1.4/connectors/vitess.html[reference documentation] to learn more.
+
+== Re-snapshot specific tables
```
gunnarmorling, line 25: I don't think this section is quite right; that new option by itself won't allow for re-snapshotting by itself: connectors would miss events between the previous shutdown and the point where the new selected snapshot is finished. The only exception would be Postgres with a custom `Snapshotter` that indeed can keep reading from the last offset, then do the snapshot, and then continue streaming reading.

This option is meant for a use case where you want to stream _all_ tables but only need an initial snapshot of some.

gunnarmorling: Rewrote this accordingly.

blog/2020-10-22-debezium-1-4-alpha1-released.adoc
```diff
@@ -0,0 +1,96 @@
...
+* New Vitess connector
+* Allow re-snapshot of specific tables
+
+Overall, the community fixed https://issues.redhat.com/issues/?jql=project%20%3D%20DBZ%20AND%20fixVersion%20%3D%201.4.0.Alpha1%20ORDER%20BY%20issuetype%20DESC[41 issues] for this release.
+Let's take a closer look at some of the highlights.
+
+== Vitess Connector
+
+https://www.vitess.io[Vitess] is a database solution for deploying, scaling, and managing large clusters of MySQL.
+We are excited that the Vitess engineers decided to build a CDC solution based on Debezium and to https://www.github.com/debezium/debezium-connector-vitess[open-source it] under the Debezium umbrella.
```
gunnarmorling, line 19: Actually it's engineers of Bolt not Vitess itself who did this :) I've updated it.

gunnarmorling:
Rebased and applied. Thx, @Naros.

PR #522, opened 2020-11-04
Setup Antora to generate documentation

uidoyen:


indraraj (approved):

indraraj:
defer merging it until docker set gets working

PR #527, opened 2020-11-09
DBZ-2732 Added Bolt to Debezium users

rgibaiev:


jpechane:
@rgibaiev Applied, thank you!

PR #533, opened 2020-11-10
Handle redirect for links in old posts

mdrillin:
Handle redirects for links in old blog posts, etc.  
- created redirects folder which contains files for all of the necessary redirects.  They just contain front matter with the permlink and a 'redir-to' variable
- Created a redirect layout which does a substitution for the "redir_to" front matter variable to put the correct version
- Fixed issues with some of the blog posts (images and redirects)
- added missing license.asciidoc


indraraj (approved):
Tested locally, we can merge it as I did get any better way to implement it.

PR #536, opened 2020-11-12
migration - updates to blog posts

mdrillin:
- Fixes to preserve formatting in blog post excerpts - defined 'excerpt_separator' in _config.yml and utilized in the blog posts.  If a blog post does not contain the separator, the excerpt is generated without html tags.
- Fix broken blog post links to other blog posts, by fixing blog post url template to agree with original site
- added one missing blog post
- removed awestruct info from blog posts and made front matters consistent in all of the posts

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -43,7 +43,10 @@ defaults:
     scope:
       type: "posts"
     values:
-      permalink: /blog/:title/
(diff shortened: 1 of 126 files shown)


indraraj (approved):
Tested locally, looks good.

PR #543, opened 2020-11-13
Update users.asciidoc

jdowning:
Add Heroku Streaming Data Connectors (managed Debezium)


gunnarmorling:
Thanks! I'd see Heroku rather in the "Integrators" category further below.
Could you add it there?

>

jdowning:
@gunnarmorling 🤦 I should have read further down. It's already listed in that section!

gunnarmorling:
Haha, so should have I, also didn't realize it was there already :) Thanks for your effort nevertheless!

PR #544, opened 2020-11-16
DBZ-1984 Move CI for website to GH Actions

ani-sha:
https://issues.redhat.com/browse/DBZ-1984

cc: @Naros


jpechane:
@ani-sha Please check this PR, it seems to me that the action was not triggered

ani-sha:
> @ani-sha Please check this PR, it seems to me that the action was not triggered

@jpechane The actions would be triggered upon landing on `develop` branch. Then it would only `build` the website in case of any `pull_request` and `build_and_deploy` in case of `push`. And we didn't had any already existing actions to be perfomed. Am I right? :thinking:

jpechane:
Yes, you are right, I'll try tomerge it after the current release

.github/workflows/build.yml
```diff
@@ -0,0 +1,13 @@
...
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build
+        run: docker run --rm --privileged=true --user=$UID:$UID -i -p 4242:4242 -e LC_ALL=C.UTF-8 -e LANG=C.UTF-8 -v $(pwd):/site -v ~/github/debezium:/debezium debezium/awestruct bash
```
Naros, line 12: This particular workflow actually doesn't do what you might think.  This particular run command actually tells docker to start the container and run the command shell `bash` but nothing else.  We do this when we run the container manually so we can start/stop the web server as needed.

I think we first need to decide are we going to base the GitHub Actions workflows on the container images or are we going to use the actions framework to do the necessary steps inline, rather than rely on the container?  If we look back at the original Travis CI configuration, we did the latter.

If we want to use the container image itself for these steps, then we actually need to review all the environment vars that the `Rakefile` needs and make sure that we're passing those to the container using `-e VAR=VALUE` syntax.  This includes things like the credentials secrets used for deployment, the `TRAVIS_PULL_REQUEST` toggle that prevents deployment for pull requests, etc.

Additionally we're going to require a new `Rakefile` task to trigger Antora, perhaps something like:
```
task :documentation => :check do
  run_antora
end
```
We would then run the docker container as `docker run .... debezium/awestruct rake clean documentation travis`.  This would run rake inside the container, it would perform a clean, build the antora documentation, followed by doing the travis task but only the build with no deployment since we'd pass `-e TRAVIS_PULL_REQUEST=1`. 

Does this make sense?

.github/workflows/build-deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+      - develop
+
+jobs:
+  build-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build
+        run: docker run --rm --privileged=true --user=$UID:$UID -i -p 4242:4242 -e LC_ALL=C.UTF-8 -e LANG=C.UTF-8 -v $(pwd):/site -v ~/github/debezium:/debezium debezium/awestruct bash
```
Naros, line 15: Similar to my comments on the `build.yml`, the same holds true here.  If we intend to use the docker container to build and deploy to GitHub, we'd need to call this container in the exact same way we do for the `build.yml` with a few minor exceptions:

1. Do not pass `TRAVIS_PULL_REQUEST=1` but instead `TRAVIS_PULL_REQUEST=0`.
2. The command you would run remains the same, `rake clean documentation travis`.

This would trigger the building of documentation & the website as well as deployment to GH pages.

.github/workflows/build-deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      - name: Deploy
```
Naros, line 17: Awestruct handles the deployment itself, so this step isn't required.  We may need to use this when we swap to Hugo, that's still TBD; so you may wish to simply comment this out of the file temporarily.

Naros (changes requested):

.github/workflows/build.yml
ani-sha: Absolutely makes sense!! We can go with running the docker container and then building antora documentation. 
Can we rename the `TRAVIS_PULL_REQUEST` var?

.github/workflows/build-deploy.yml
ani-sha: Will be commenting those for now! Thnx.

.github/workflows/build.yml
ani-sha:
```
docker run --rm --privileged=true --user=$UID:$UID -i -p 4242:4242 -e LC_ALL=C.UTF-8 -e LANG=C.UTF-8 -e TRAVIS_PULL_REQUEST=1 -v $(pwd):/site -v ~/github/debezium:/debezium debezium/awestruct rake clean documentation travis
``` 
@Naros Running out the docker command in GitHub Actions throws some error. 
```
Status: Downloaded newer image for debezium/awestruct:latest
Could not find Platform-0.4.2 in any of the sources
Run `rake setup` to install required gems from RubyGems.
```

debezium-builder: > Can we rename the `TRAVIS_PULL_REQUEST` var?

Absolutely, we can certainly rename it we just need to also then adjust the `Rakefile` accordingly too.

> Could not find Platform-0.4.2 in any of the sources
Run `rake setup` to install required gems from RubyGems.

I believe Travis CI automatically saw the gem files because we setup the YAML based on `Ruby`.  With the awestruct image, you may need to run the command as `bundle update && rake clean documentation travis` so that the ruby gems are installed properly.

Naros: > Can we rename the `TRAVIS_PULL_REQUEST` var?

Absolutely, we can certainly rename it we just need to also then adjust the `Rakefile` accordingly too.

> Could not find Platform-0.4.2 in any of the sources
Run `rake setup` to install required gems from RubyGems.

I believe Travis CI automatically saw the gem files because we setup the YAML based on Ruby. With the awestruct image, you may need to run the command as bundle update && rake clean documentation travis so that the ruby gems are installed properly.

ani-sha: Okay. Btw @gunnarmorling suggested halting this until the website migration. So that it would be easier to build directly with Jekyll.

gunnarmorling: Yes, after seeing how easy it is with Jekyll (see https://github.com/quarkusio/quarkusio.github.io/blob/develop/.github/workflows/deploy.yml), I think we should just wait for the switch to Jekyll next week.

gunnarmorling: Ok, so here's what I think we should do:

- create a deploy key with write access (@jpechane can help with that)
- create a GH Action very similar to [this one](https://github.com/gunnarmorling/morling.dev/blob/master/.github/workflows/gh-pages-deployment.yml), only that
  - instead of calling Hugo, it'd call `rake build`, using the _debezium/website-builder_ container image (this still needs to be deployed to DockerHub, via @jpechane)
  - it'd use PUBLISH_BRANCH: master and PUBLISH_DIR: ./_site
- remove the "travis" task from the Rakefile in the website repo
- rename the "deploy" task from the Rakefile to build and have it besides running Antora also call `JEKYLL_ENV=production bundle exec jekyll build` (as per the [Jekyll docs](https://jekyllrb.com/docs/step-by-step/10-deployment/))

I.e. build the website via Jekyll through our container image and then use the standard GH Pages GitHub Action for deploying it.

ani-sha:
Closing.
Refer https://github.com/debezium/debezium.github.io/pull/585

PR #546, opened 2020-11-16
Website migration updates

mdrillin:
- now sorting tags on blog page to agree with current site
- adding recent release content


indraraj (approved):
Tested locally, looks good.

PR #549, opened 2020-11-18
Removed the added versions.yml files and updated the release code

indraraj:


mdrillin:
If you look at 'see older versions' on the releases page, the 0.9 and 0.10 versions are in the wrong order.  Jekyll strips the dot then orders alphabetically, which yields the wrong order.  Not sure atm how this can be solved, I'll look into it further

mdrillin:
@indraraj - upon looking at this further, I think I'd just leave the '_data/versions.yml' file (unless you can think of a clever way to address the ordering issue mentioned in previous comment).  '_data/versions.yml' would rarely need to be changed, so I don't see it as a big problem. Addition of 'version' to the other .yml files looks good...  wdyt?

indraraj:
> If you look at 'see older versions' on the releases page, the 0.9 and 0.10 versions are in the wrong order. Jekyll strips the dot then orders alphabetically, which yields the wrong order. Not sure atm how this can be solved, I'll look into it further

ohk I see what is happening here.

indraraj:
> @indraraj - upon looking at this further, I think I'd just leave the '_data/versions.yml' file (unless you can think of a clever way to address the ordering issue mentioned in previous comment). '_data/versions.yml' would rarely need to be changed, so I don't see it as a big problem. Addition of 'version' to the other .yml files looks good... wdyt?

Hey, @mdrillin tried to achieve it today, but no success. What is happening here is that there is no default sorting of the file inside `_data`, it returns somewhat like `ls -U` response.( https://stackoverflow.com/questions/48345864/jekyll-what-is-the-default-data-sorting-criteria?answertab=active#tab-top ) And our sorting is also not useful because till that time "." have been removed from the file name. 

@mdrillin  I second your opinion here. Since it would be just one file(at the top-level `_data/releases`) it may not increase much complexity in the release process. We can once confirm with @Naros on this.

mdrillin:
@Naros @gunnarmorling - are you guys ok with leaving one '_data/versions.yml' file which would need to be updated with each <major>.<minor> release?  We could potentially get this removed at some point, but jekyll has not been cooperating in getting the current _data directory structure sorted properly.  We're proposing to defer this until after cutover

gunnarmorling:
> are you guys ok with leaving one '_data/versions.yml' file which would need to be updated with each major.minor release?

@mdrillin, yes, that's ok. We can take that one separately. Does this mean this PR needs some update though, as I see this file being deleted. It needs a rebase in any case, @indraraj.

gunnarmorling:
An update to https://github.com/debezium/debezium/blob/master/RELEASING.md might be warranted (CC @jpechane).

indraraj:
> > are you guys ok with leaving one '_data/versions.yml' file which would need to be updated with each major.minor release?
> 
> @mdrillin, yes, that's ok. We can take that one separately. Does this mean this PR needs some update though, as I see this file being deleted. It needs a rebase in any case, @indraraj.

yes, I was waiting for confirmation to update it. I will update it now. @gunnarmorling

mdrillin (approved):
retest looks good - will merge

PR #550, opened 2020-11-19
Updated the css to change the nav bar collapse toggle point and Fixed…

indraraj:
… the LTS version link in banner

- Fixed the LTS version in banner.

- Updated the css to change the toggle point for nav menu collapse.


mdrillin (approved):
tested locally, lgtm

PR #555, opened 2020-11-19
Add support for Google Analytics

uidoyen:


_layouts/base.html
```diff
@@ -35,7 +24,10 @@
   </div>
   {% include project-footer.html %}
   {% include redhat-footer.html %}
-    
+  
+  {% if site.google_analytics %}
```
gunnarmorling, line 28: So this would _always_ add this, right? Whereas the snippet shouldn't be included when running local for dev, or on staging in the future. Looks like there a notion of "environments" or similar in Jekyll, which we could query here. Look for "jekyll.environment == 'production'" in https://desiredpersona.com/google-analytics-jekyll/.

gunnarmorling (commented):
Good start, one comment on the environment issue, @uidoyen.

_layouts/base.html
uidoyen: > Good start, one comment on the environment issue, @uidoyen.

Added it for production environment.

gunnarmorling:
LGTM now, but seems to need a rebase. Will leave it to you folks to merge it. Thanks!

mdrillin (approved):
looks good

Issue #551, opened 2020-11-19
Move redundant footer from posts to template


gunnarmorling:
Atm. all the blog post AsciiDoc files have the same footer ("About Debezium", "Get Involved"). This should be extracted into the template. Might also go _under_ the author info actually.

@mdrillin @indraraj @uidoyen Here's a first follow-up improvement.

mdrillin:
pending PR #558

mdrillin:
merged, closing

PR #558, opened 2020-11-19
moves repeated blog section into template

mdrillin:
Addresses issue #551 

- add 'About Debezium' and 'Get involved' sections into the blog post template, underneath the author block
- remove the repeated section from all of the blog post asciidocs

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -122,6 +122,33 @@ layout: base
       
     
     {% endif %}
+    <div class="row">
+      <div class="col-md-12">
+        <hr>
+        <h2>About Debezium</h2>
+        <p>
+          Debezium is an open source distributed platform that turns your existing databases into event streams,
+          so applications can see and respond almost instantly to each committed row-level change in the databases.
+          Debezium is built on top of <a href="http://kafka.apache.org/">Kafka</a> and provides <a href="http://kafka.apache.org/documentation.html#connect">Kafka Connect</a> compatible connectors that monitor specific database management systems.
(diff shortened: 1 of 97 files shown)


_layouts/post.html
```diff
@@ -122,6 +122,32 @@ <h2>{{ authorName }}</h2>
       
     
     {% endif %}
+    <div class="row">
```
indraraj, line 125: This need to be moved above the "about author card"

indraraj:
Tested locally, looks good. Just need to place it above the author card to match the current site.

gunnarmorling:
In fact I had suggested to put it below the author card as right now there's no clear boundary between the actual post and that standard "footer". How does it look, perhaps you could attach a screenshot?

indraraj:
image

 @gunnarmorling

indraraj:
sry, I missed reading the complete description of the Github issue. I see that it is something you actually asked for, you can go ahead and merge this @mdrillin

indraraj (approved):

gunnarmorling:
Thanks, so what do you which one is better? Perhaps the two footer paragraphs should be styled slightly different than the main text?

indraraj:
I think it's good to move it below the "About author" card as it's not part of the blog content. Yes, different styling will be great but for now, we can put some blog end marker/separator. wdyt?

image

gunnarmorling:
Yes, good idea. 👍

mdrillin:
Updated the commit to include separator - merging.

Issue #563, opened 2020-11-20
Order versions drop-down from new to old


gunnarmorling:
See selector at upper right on http://localhost:4000/releases/1.3/; let's put the newest first, matching what's done by Antora in the docs.

@mdrillin @uidoyen @indra

indraraj:
Pr raised

indraraj:
pr merged.

PR #570, opened 2020-11-23
Updated the version drop-down order in release breadcrumb

indraraj:


uidoyen (approved):
Tested locally. Looks good.

mdrillin:
@indraraj  -  why was the ordering reversed here?  It is a deviation from the original site, did someone request the change?

versionReversal

gunnarmorling:
> did someone request the change?

Yes, I did :) See https://github.com/debezium/debezium.github.io/issues/563 (good idea btw. to always reference the corresponding issue in commit messages). The motivation being to have the newest one always at the top (also consistent with the ordering in the docs).

mdrillin:
Ok, makes sense - thanks!

PR #579, opened 2020-11-23
apply PR504 for website-migration

mdrillin:
applying the change from the referenced PR.  @uidoyen It looks like we may have missed this change in website-migration branch - can you confirm?  https://github.com/debezium/debezium.github.io/pull/504


uidoyen:
@mdrillin You are right. We have missed it. It looks good now. Thanks.

uidoyen (approved):
LGTM.

PR #594, opened 2020-11-25
Clearing up unnecessary files and folders

uidoyen:
- Removed _SCSS folder
- Removed Fontawesome fonts from folder
- Changed bootstrap from CDN to local directory
- Removed Fontawesome CDN
- Removed mobile-nav.js (It is not required)
- Removed TitilliumText22LBold from custom.css
- Removed scss from config


mdrillin (approved):
LGTM - pulled down and tested ok

Issue #595, opened 2020-11-25
On specific version release page, documentation button link is incorrect


mdrillin:
On the specific version release page, the documentation button always links to the latest stable version of the documentation.  It should link to the specific version, not the latest stable

![DocumentationButton](https://user-images.githubusercontent.com/1820403/100253467-c5d03d00-2f06-11eb-856c-37bc0a53a402.png)

mdrillin:
pending PR #596

gunnarmorling:
PR merged. Thx!

PR #596, opened 2020-11-25
cleanup redirect page header, fix release version page issue

mdrillin:
- cleanup of redirect.html, same as in pending @uidoyen PR
- fix issue with release version page documentation link, issue #595


gunnarmorling (approved):
Good catch, thx!

Issue #599, opened 2020-11-26
Disqus comments aren't loaded


gunnarmorling:
I pushed 9540910323db2e7318f680a7ec9b0a18973dcc1a as a work-around, hard-coding the base URL into the Disqus include for now.

That way, the comment widgets are loaded again, but ideally we won't have the base URL in any template/include. With Awestruct, the base URL was part of the environment-specific configuration and could be referenced in templates as a variable then. Is something similar doable for Jekyll?

@mdrillin @indraraj @uidoyen

uidoyen:
@gunnarmorling Yes, It is doable. We can define `url:"https://debezium.io/"` in `_config.yml` and then it will be accessible as `{{site.url}}`

gunnarmorling:
Sweet, thx! PR merged.

PR #604, opened 2020-11-27
Adding base URL to config and Disqus

uidoyen:


gunnarmorling (approved):
LGTM; we might want to extract that into a separate (environment-specific) config file at some point (see https://jekyllrb.com/docs/configuration/environments/), but for now that's good. Thx!

gunnarmorling:
Ah, just noticing a bit too late: please prefix commit messages with the issue key, e.g. "#599..."

PR #605, opened 2020-11-27
redirect with trailing slash

indraraj:


gunnarmorling:
@indraraj, can you please always reference the issue number in your commit messages?

gunnarmorling:
Applied, thx!

PR #623, opened 2021-01-13
DBZ-2907 Adding Vitess Connector

uidoyen:
Added Vitess Connector plug-in and updated `series.yml` to make download connector plug-in drop-down data-driven.

### DIFF
--- a/_data/releases/0.1/series.yml
+++ b/_data/releases/0.1/series.yml
@@ -3,4 +3,5 @@ displayed: false
 hidden: true
 tutorial: false
 connectors:
-  - mysql
+  - id: mysql
+    title: MySQL Connector Plug-in
--- a/_data/releases/0.1/series.yml
+++ b/_data/releases/0.1/series.yml
@@ -3,5 +3,4 @@ displayed: false
 hidden: true
 tutorial: false
 connectors:
-  - id: mysql
-    title: MySQL Connector Plug-in
+  - mysql
--- a/_includes/release-version.html
+++ b/_includes/release-version.html
@@ -10,6 +10,9 @@
 </div>
 <h3 class="section" id="compatibility">Tested Versions</h3>
 
+
+
+
 {% assign formattedVersion = page.debezium-version | replace: ".", "" %}
 {% assign releaseDetails = site.data.releases[formattedVersion] %} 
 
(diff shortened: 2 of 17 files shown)


_includes/release-version.html
```diff
@@ -109,49 +109,24 @@ <h5 class="version-date">
...
-              >
-            </li>
-            <li>
-              <a
-                href="https://repo1.maven.org/maven2/io/debezium/debezium-connector-db2/{{releaseDetails[subVersion].version}}/debezium-connector-db2-{{releaseDetails[subVersion].version}}-plugin.tar.gz"
-                >Db2 Plug-in</a
-              >
-            </li>
+            {% for connector in releaseDetails.series.connectors %}
+              {% if connector.id == 'postgresql' %}
```
gunnarmorling, line 115: Can't we use "postgres" as the id then to avoid this special handling?

_data/releases/0.1/series.yml
```diff
@@ -3,4 +3,5 @@ displayed: false
 hidden: true
 tutorial: false
 connectors:
-  - mysql
+  - id: mysql
+    title: MySQL Connector Plug-in
```
gunnarmorling, line 7: Instead of adding this to each series file, can we have one file `_data/connectors.yml`, which contains the id/title pairs of all connectors. Then in the series files, you keep it to the id only, as before. In the template, you resolve the readable name using the id against that connectors file. WDYT?

_includes/release-version.html
uidoyen: Yes, We can use `postgres` as it's not being used anywhere else apart from release pages.

_data/releases/0.1/series.yml
uidoyen: I think its a good approach. Updated accordingly.

keweishang:
Hi @gunnarmorling @uidoyen, Vitess connector supports Vitess version `8.0.x` as described in the https://debezium.io/documentation/reference/connectors/vitess.html

> Vitess version 8 is supported

Could you also change the `_data/releases/1.4/series.yml`'s corresponding part to:
```
  vitess:
    database:
      versions:
        - 8.0.x (this is the change)
    driver:
      versions:
        - 7.0.0 (no change for this, we're still using vitess-grpc-client version 7)
```

I can create a separate ticket for it or add this in the existing ticket's description.

gunnarmorling:
> Could you also change the _data/releases/1.4/series.yml's corresponding part to:

Good catch, thx! Fixed, will be live in a bit.

PR #628, opened 2021-01-21
Changing environment to production

uidoyen:


gunnarmorling:
I think we should leave the default to "development", and rather adjust the GH Action workflow to build with the production "profile".

uidoyen:
> I think we should leave the default to "development", and rather adjust the GH Action workflow to build with the "production" profile.

I looked into this, we can use GH Action workflow to build the site but `jekyll-minifier` plugin throw an error. I think GH Action doesn't support this plugin.

gunnarmorling:
I meant that the existing workflow needs adjustment :) But no worries, I've just done it:

https://github.com/debezium/debezium.github.io/compare/5b5b1d73462b233e3329db1be659b57e98ee53dd...e8f8bca39dbddf0bd5c861ad35ff40ef076edab1

Going to close this PR therefore.

PR #629, opened 2021-01-27
DBZ-2978 Adding Syntax Highlighter(hljs) 

uidoyen:


gunnarmorling:
Excellent, thx! Can you push this to staging, please?

uidoyen:
> Excellent, thx! Can you push this to staging, please?

Already pushed on staging https://debezium-builder.github.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/

gunnarmorling:
Ok, nice overall. There's one issue with call-outs still. E.g. look here:


https://debezium-builder.github.io/blog/2019/10/01/audit-logs-with-change-data-capture-and-stream-processing/

In the second listing, (1), (2), etc. should be rendered using those bubble
icons, as (4), (5), (6) below the listing. Same for the explanations 1 - 3
for those call outs. It works for 4 - 6, as I manually added them there,
instead of relying on the default call-out mechanism (look at the AsciiDoc
source to see what I mean).

>

uidoyen:
Pushed on staging for testing https://debezium-builder.github.io/blog/2019/10/01/audit-logs-with-change-data-capture-and-stream-processing/ and updated the PR

assets/javascript/highlight.min.js
```diff
@@ -0,0 +1,3 @@
+/*! highlight.js v9.11.0 | BSD3 License | git.io/hljslicense */
```
gunnarmorling, line 1: Is this file needed actually? All highlighting happens on the server side, right? IF we don't needed, let's remove it, avoiding any confusion.

gunnarmorling (approved):
LGTM, call-outs rendered correctly now. One question line. Thanks!

assets/javascript/highlight.min.js
uidoyen: I think we need this. It highlight keywords, differences can be seen below.
without `highlight.js`
Screenshot from 2021-01-28 17-35-20

with `hightlight.js`
Screenshot from 2021-01-28 17-34-53

gunnarmorling:
Hum, hum, interesting. So is this used at build time then? I.e. with that
script being present, if you have JS disabled in the browser, will the
outcome be the same as with JS enabled?

>

uidoyen:
> Hum, hum, interesting. So is this used at build time then? I.e. with that script being present, if you have JS disabled in the browser, will the outcome be the same as with JS enabled?
> […](#)

I think it find some matching tags and add `css` to it. Outcome will be little different if JS is disabled in the browser. It might look like first screenshort(without `highlight.js`)

uidoyen:
We might show some warning if JS is disabled. WDYT?

gunnarmorling:
@uidoyen, taking a step back. What was behavior before the migration; was high-lighting purely server-side back then? I pushed tags `develop-website-pre-migration` and `master-website-pre-migration` back then, so you could take a look. If it was purely server-side (which I always assumed), I'd prefer to stick to that.

uidoyen:
> @uidoyen, taking a step back. What was behavior before the migration; was high-lighting purely server-side back then? I pushed tags `develop-website-pre-migration` and `master-website-pre-migration` back then, so you could take a look. If it was purely server-side (which I always assumed), I'd prefer to stick to that.

I checked the tags, actually it wasn't purely server-side earlier, it was using `highlight.js` as well.

gunnarmorling:
Ok, gotcha! Will apply then. Thanks!

PR #636, opened 2021-02-08
Debezium 1.5 Alpha1 announcement

gunnarmorling:


Naros (approved):
You caught my only issue with the original push so now it LGTM.

gunnarmorling:
> You caught my only issue with the original push so now it LGTM.

Ha, excellent :) Thx!

>

PR #638, opened 2021-02-09
DBZ-3076 DB driver label change for non-relational connectors

uidoyen:


gunnarmorling (approved):
Ok, not a huge fan of hacking these things into the UI, but it should do the trick. Thx!

PR #685, opened 2021-06-29
Adding separate field to provide summary for release series

indraraj:
@gunnarmorling  knowing the char limit(120) I tried to keep the summary sort and in one line. 

And if we wanted to have something like 2 or 3 lines description. I would suggest we keep the existing 120 char inside the box but also show the complete version on the detail page(screen below). At the top just above `Tested versions`
image

### DIFF
--- a/_data/releases/0.10/series.yml
+++ b/_data/releases/0.10/series.yml
@@ -1,5 +1,6 @@
 summary: Version 0.10 release stream
 version: "0.10"
+overview: New incubating Cassandra connector, new features and enhancements for PostgreSQL connector, and various bugfixes
 displayed: false
 hidden: false
 tutorial: true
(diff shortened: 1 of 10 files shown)


gunnarmorling:
Hey @Naros, @jpechane, wanted to point you to this change; there'll be a new attribute taking a description of the contents of a release family (1.4, 1.5, 1.6, etc.). So far, we only showed the contents of the _latest_ micro for reach series, which typically would result in rather unspecific descriptions like "bugfix release". @indraraj already changed this to show the _first_ one (Alpha1), which changes the focus towards more feature-related changes, but it's missing all the other stuff added in subsequent preview releases. So we decided to go for a new attribute at the series level. So when doing a release going forward, this will need some attention (defining an initial value for Alpha1, updating the value when doing subsequent preview releases that contribute significant features). Hope that makes sense, let's discuss if there are any questions.

gunnarmorling (approved):

gunnarmorling:
Rebased and applied. Thanks a lot, @indraraj!

PR #695, opened 2021-07-20
Update users.asciidoc

marcelojscosta:
Updating Users ASCII Doc to include some companies in Brazil adopting Debezium


gunnarmorling:
Woot, thanks a lot for these additions, @marcelojscosta! Could you only rework the commit so keep the list in alphabetic order; and if there's no public reference to that Puma2 project, I'd suggest to just keep the company name, as that info isn't too helpful for readers then anyways. Thanks again!

gunnarmorling:
Reordered alphabetically and applied. Thx, @marcelojscosta!

PR #721, opened 2021-09-23
DBZ-4011 Surge preview GitHub actions fix

uidoyen:


gunnarmorling:
Wow, this became quite complex. There's no simpler way indeed? I didn't quite expect that from looking at the description of that preview action 🤔 .

PR #734, opened 2021-10-13
DBZ-3897 Add rake method to copy specified versions

uidoyen:
https://issues.redhat.com/browse/DBZ-3897


uidoyen:
@gunnarmorling updated accordingly.

gunnarmorling:
@uidoyen, we're getting there. Can you:

* Make it "Debezium 1.7 Documentation" rather than "Debezium Documentation 1.7"
* Add the version to the main title page (as shown in the screenshot above)
* Have the nightly version exposed via https://debezium-debezium-github-io-preview-pr-734.surge.sh/documentation/reference/nightly/index.html
* Add a "dev" URL path, which at this point in time would be sourced from master (i.e. it's the same like "nightly"), but after the first 1.8 release (Alpha1) would be sourced from 1.8
* Make the links "Go to stable" in the note boxes for nightly and older releases point to the "stable" path rather than 1.7

Thanks!

Rakefile
```diff
@@ -86,6 +90,28 @@ task :author do
...
+  latestStableVersion = playbook['asciidoc']['attributes']['page-version-current']
+  latestDevelVersion = playbook['asciidoc']['attributes']['page-version-devel']
+
+  stableDir = "_site/documentation/reference/#{latestStableVersion}";
+  develDir = "_site/documentation/reference/#{latestDevelVersion}"
+
+  if File.exists?(stableDir)
+   FileUtils.cp_r stableDir, "_site/documentation/reference/stable"  
+  else
+    puts "Unable to find stable version dir"
```
Naros, line 124: Should this and the occurrence below where we copy to the `devel` directory throw an exception?  I'm just curious if there is some reason during website publishing by GitHub actions if we should somehow trigger the action to fail?

_antora/supplemental_ui/partials/banner.hbs
```diff
@@ -1,12 +1,12 @@
...
     This partial is responsible for rendering one of 3 banners at the top of each article:
-
         banner-current.hbs
         banner-development.hbs
         banner-outdated.hbs
 --}}
-{{#if (or (eq page.version 'master') (gt page.version page.attributes.version-current))}}
-    {{!-- For antora.yml that references master or a version less than current defined in playbook, add development banner --}}
+
+{{#if (or (eq page.version 'nightly') (gt page.version page.attributes.version-current))}}
```
Naros: Would it make sense to compare the version to `page.version.devel` rather than the hard-coded `nightly` value?

Naros (commented):
Few comments inline.

_antora/supplemental_ui/partials/banner.hbs
uidoyen: @Naros thanks! Updated accordingly.

Rakefile
uidoyen: @Naros Currently its just handling Antora side but I think Its a probably good to have during website publication by GitHub action.

Naros:
@uidoyen could you also rebase the PR when you get a chance?

gunnarmorling:
@uidoyen, is there a preview of this anywhere? Seems the message above didn't get updated somehow.

uidoyen:
> @uidoyen, is there a preview of this anywhere? Seems the message above didn't get updated somehow.

@gunnarmorling had some issue with GitHub server. It's available now. https://debezium-debezium-github-io-preview-pr-734.surge.sh

Naros:
I have a few outstanding questions

- What should be the canonical URL for the nightly (/) documentation, `/stable/` or `/1.7/`?
- What should be the canonical URL for the stable (/stable) documentation, `/stable/` or `/1.7/`?
- What should be the canonical URL for the older releases (i.e. /1.6/), `/stable/` or `/1.7/`?

I propose these questions because from what I can tell in the preview is the answer to all the above is currently `/1.7/` and I just want to make sure we're all on the same page that's what we wanted.

gunnarmorling:
Good questions, @Naros! Canonical URLs should generally be the one under `/stable`. I.e. that's the ones to which all others should point. Because they are the ones we'd one to share in links e.g. on the ML and which should rank best in Google search results.

As far as the URL structure itself looks like, and also in regards to the values in the drop-downs, I think this is looking good now. The only wish I had would be to get the version number into the main page title: "Debezium 1.7 Documentation". This would be in particular useful for those users arriving to the docs through one of the aliases (nightly, devel, stable), as they then will immediately know which version this is about.

Naros:
> Canonical URLs should generally be the one under /stable. I.e. that's the ones to which all others should point. Because they are the ones we'd one to share in links e.g. on the ML and which should rank best in Google search results.

So looking at what is currently committed in the main Debezium repo here and based on the above, I think there is one outstanding change we need to make to the antora.yml files in the main repository:

In the **1.7** branch, we currently only define the `version: 1.7` attribute which means when canonical URLs get resolved its going to use that value.  Since we want those to actually reference `/stable/` we should adjust the configuration in the `antora.yml` file to be like the following:
```
version: `stable`
display_version: `1.7`
```
This change should then also adjust the master and older branches to point to the `/stable/` canonical URL rather than `/1.7/`.

This would mean that rather than copying `/1.7/` to `/stable/` we'd probably need to invert that copy logic to then copy the contents of the `/stable/` path to the `/1.7/` path.

uidoyen:
@gunnarmorling rebased!

gunnarmorling:
Squashed and applied. Thx, @uidoyen!

PR #745, opened 2021-10-28
DBZ-3626 Renaming master branches to gh-pages/main

Naros:
https://issues.redhat.com/browse/DBZ-3626


gunnarmorling:
@uidoyen, this one is a bit odd: the preview env has _not_ been destroyed, still the comment above was added. Something wrong with checking return values of the Surge execution perhaps? I am going to tear down that env manually now.

PR #753, opened 2021-11-12
DBZ-4214 GH Action for tearing down abandoned website preview

uidoyen:
https://issues.redhat.com/browse/DBZ-4214

### DIFF
--- a/.github/workflows/surge_preview_teardown_manually.yml
+++ b/.github/workflows/surge_preview_teardown_manually.yml
@@ -4,13 +4,18 @@ on:
   workflow_dispatch:
     inputs:
       PRNumber:
-        description: 'PR number to teardown'     
+        description: 'PR number to teardown'
         required: true
 
 jobs:
-  preview:
+  success:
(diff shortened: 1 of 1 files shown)


gunnarmorling:
LGTM, but this part discussed in the issue still seems to be missing:

> This action should also log out the URLs of all the currently existing preview environments. 

I'd suggest to 

* first tear down the specified environment, then display the remaining ones (if any)
* if no environment has been specified, simply display any existing ones (if any)

uidoyen:
@gunnarmorling Updated accordingly.

gunnarmorling:
Applied, thx!

gunnarmorling:
Ok, so this generally works, thanks. Could you only make the PR number optional and skip the tear down step if none is given, only displaying the existing environments? Also adjust the parameter description to something like "PR number to tear down; omit to list all existing preview environments". No need for another PR review cycle for that, please just merge that change yourself and resolve the issue in Jira after that. Thx!

PR #774, opened 2022-01-21
DBZ-3953 Publish tested postgres plugins

jpechane:
https://issues.redhat.com/browse/DBZ-3953


gunnarmorling:
@jpechane, LGTM. Could you only rebase it and add the info for 1.9, now that the Alpha went out?

jpechane:
@gunnarmorling Done

PR #776, opened 2022-01-31
Add Airwallex to users.asciidoc

brettjordan:
Here you go Gunnar! When we have a blog post about our Debezium usage I will send another PR with details.


gunnarmorling:
Applied, thx!

> When we have a blog post about our Debezium usage I will send another PR with details.

Perfect, looking forward to it. This could also be added to _documentation/online-resources.asciidoc_ then.

PR #791, opened 2022-04-13
Fix link in page footer to install page

ciwchris:


gunnarmorling:
Good catch, thanks! While at it, I'll also change all these links in the footer to point to the "stable" docs version.

PR #805, opened 2022-06-01
Add Shippeo to Debezium users + online resources

omarghalawinji:
Hello,
We wrote a blog post on Shippeo's experience using Debezium to replicate data to Snowflake and we'd like to add it to the online resources.
Thanks!


jpechane:
@omarghalawinji Applied, thanka lot for the blogpost. it is really full of gems!

omarghalawinji:
Thank you @jpechane !

PR #818, opened 2022-07-13
DBZ-5048 Rename GH repository with container images

vjuranek:
https://issues.redhat.com/browse/DBZ-5048

There are still some mentions of `docker-images` in the old documentations, but it would mean to rewrite git history in the main repo, so there's probably nothing we can do about it. Besides documentation I find anything else but the places fixed in this PR.

### DIFF
--- a/_posts/2017-02-08-Support-for-Postgresql.adoc
+++ b/_posts/2017-02-08-Support-for-Postgresql.adoc
@@ -32,6 +32,6 @@ In general, there are three major steps involved in getting the connector runnin
 2. Setting up the PostgreSQL server with appropriate replication permissions
 3. Starting the Kafka Connect, Broker and Zookeeper machines
 
-For steps 1 and 2 you can check out our https://github.com/debezium/docker-images/tree/master/postgres/9.6[PostgreSQL Docker image] together with the sources for the https://github.com/debezium/postgres-decoderbufs[logical decoding plugin]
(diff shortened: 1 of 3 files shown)


_posts/2017-02-08-Support-for-Postgresql.adoc
```diff
@@ -32,6 +32,6 @@ In general, there are three major steps involved in getting the connector runnin
 2. Setting up the PostgreSQL server with appropriate replication permissions
 3. Starting the Kafka Connect, Broker and Zookeeper machines
 
-For steps 1 and 2 you can check out our https://github.com/debezium/docker-images/tree/master/postgres/9.6[PostgreSQL Docker image] together with the sources for the https://github.com/debezium/postgres-decoderbufs[logical decoding plugin]
+For steps 1 and 2 you can check out our https://github.com/debezium/container-images/tree/master/postgres/9.6[PostgreSQL Docker image] together with the sources for the https://github.com/debezium/postgres-decoderbufs[logical decoding plugin]
```
jpechane: @vjuranek The link description containword `Docker`. It should be removed here and probably in other parts of the doc if possible.

vjuranek: I fixed it here and in the other link name, but not sure if it worth to fix it in the rest of the blog post and other posts (`Docker` is mentioned in many other posts and release notes). Of course now we should use neutral `container` and similar instead of `Docker`, but in the past we used Docker so I don't see much value in rewriting old posts - it reflect the time when the post was written when Docker was the only viable container solution (if I don't count LXC) and IMHO it's completely fine to keep it as it is.

jpechane: @vjuranek Makes sense, thanks. Applied.

PR #822, opened 2022-07-28
We just migrated our data pipeline to debezium

anuraagheisenberg:
We used to use AWS DMS as managed solution.
Now we use strimzi kafka with debezium kafka connectors for mongo/mysql


Naros:
That's awesome news @anuraagheisenberg, thanks for the PR!.

PR #841, opened 2022-10-20
Added Myntra as a User

dyaneshvarun:
Myntra heavily uses Debezium for ingesting CDC Data. 

Added Myntra as a User -- Update community/users.asciidoc


jpechane:
@dyaneshvarun Nice, thanks a lot!

PR #847, opened 2022-11-04
Update online-resources.asciidoc

gunnarmorling:
Hey @jpechane, @Naros, so how do you want this to do going forward? Are you happy with me committing changes like this myself, or should I go through pull requests. Both is cool for me, just let me know. Thanks!

### DIFF
--- a/documentation/online-resources.asciidoc
+++ b/documentation/online-resources.asciidoc
@@ -47,6 +47,7 @@ Thanks!
 
 == Blog Posts & Articles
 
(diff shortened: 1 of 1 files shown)


jpechane:
@gunnarmorling Just commit it yourself - I don't see any added value from our review.
But still it would be good if you could create the PR and merge it yourself - just for us to get notification about the new content so we can check the articles. Would it be ok for you?

gunnarmorling:
@jpechane, sure that sounds good. Thanks!

PR #848, opened 2022-11-08
Roadmap update

Naros:
Hi @jpechane and @rk3rn3r these are the roadmap updates to carry us into 2023.  While I've mostly discussed these all with @jpechane, could you take a look @rk3rn3r and let me know if we should adjust any of these for the UI?  I've aligned these based on our internal discussions from earlier this week.


jpechane:
@Naros Applied, thanks a lot!

PR #849, opened 2022-11-09
Add 1.9 to landing page

Naros:


gunnarmorling (commented):
Interesting; I reckon that's just a stop gap solution until it can be done in a data-driven way?

Naros:
> Interesting; I reckon that's just a stop gap solution until it can be done in a data-driven way?

It is.  We need to understand moving forward, say once 2.1 or 2.2 are released, what do we expect to be visible on the landing page?  Would it continue to be "1.9", "2.x", and optionally the "2.y" (development version)?  Or would the 1.9 reference eventually go away and we fall back to just showing the latest "2.x" stable and optionally the "2.y" development version?

@jpechane what's your take on this?

Naros:
Going to merge for the time being and we can continue the discussion on how best to deal with this long-term.

jpechane:
Well, I am not exactly happy with the current solution. IIRC our intent always was to have only one stable version - even advertised. I can imagine this is an exception eith 1.x -> 2.x transition but in my opinion no later than with 2.1 release the 1.9 and 2.0 should become non-advertised versions as the others are.

PR #860, opened 2022-12-15
DBZ-5900: Fix the Debezium UI doc link visibility

indraraj:
Screenshot:
Debezium UI link is now visible in the bottom left corner 
image


jpechane:
@indraraj Applied, thanks

PR #869, opened 2023-01-06
[docs] Updating tested Postgres versions

gunnarmorling:
@Naros, I think we've tested with PG 15 at least since 1.9?


jpechane:
@gunnarmorling Applied, thanks

PR #871, opened 2023-01-18
Add Artie Transfer as an integration

Tang8330:
Artie Transfer sits downstream of Debezium and leverages Debezium to capture CDC to provide a real-time replication to DWHs


jpechane:
@Tang8330 Applied, thanks for letting us know!

PR #894, opened 2023-03-10
doc: Add Okta (Auth0) to `users.asciidoc`.

btiernay:
Adds Okta to the list of users of Debezium to https://debezium.io/community/users per [Data Pipelines on Auth0’s New Private Cloud Platform](https://auth0.com/blog/data-pipelines-on-auth0-s-new-private-cloud-platform/):

image


jpechane:
@btiernay Applied, thanks for the reference.

PR #895, opened 2023-03-15
add Gorgias to the list of debezium users

gmouss:
add Gorgias to the list of debezium users


jpechane:
@gmouss Applied, thanks!

PR #935, opened 2023-08-10
DBZ-6775 Add release note on schema changes per shard

twthorn:
Add notes on breaking changes with schema changes being applied per shard. Include guidance on how to migrate smoothly with no breaking changes/undesired topic output that uses newly added topic naming strategy in https://github.com/debezium/debezium-connector-vitess/pull/151


jpechane:
@twthorn Applied, thanks!

PR #963, opened 2023-10-10
Add a more info link into the feel for truncated content

emmanuelbernard:
When using a RSS reader, it is unclear that the content of the blog is longer than the excerpt. A classical way to make it clearer is to add a read more link. This is what this PR is doing. 
I have not been able to test it as the local docker run fails for me (amd64 vs arm64 issue).

 Note, I personally hate when blogs create excerpts as it ruins the RSS reader experience. So best would be to remove the exerpt block and always show the full content IMO.

### DIFF
--- a/blog.atom
+++ b/blog.atom
@@ -23,7 +23,8 @@ layout: none
 					<dc:creator>{{ post.author.name | xml_escape }}</dc:creator>
 				{% endif %}
 				{% if post.excerpt %}
-					<description>{{ post.excerpt | xml_escape }}</description>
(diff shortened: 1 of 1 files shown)


Naros (approved):
Thanks, @emmanuelbernard, I'm going to go ahead and add this as a quick fix.  I'll log a Jira issue to come back to this and review RSS in more detail as time permits to display the content entirely instead.

Naros:
Logged https://issues.redhat.com/browse/DBZ-7027 as follow-up.

PR #965, opened 2023-10-11
Fix blog feed - Fix liquid string concatenation

emmanuelbernard:
@Naros sorry looks like Liquid string contatenation is not how I imagined it woudl work. This should be better but you can check by running locally and hitting `/blog.atom` The end of the description should have `Read more...`.


Naros (approved):
LGTM, thanks @emmanuelbernard.

PR #972, opened 2023-11-10
Fix atom feed with blog link, not feed link

emmanuelbernard:
@Naros I found a bug in the read more link, fixed with this.


Naros (approved):
Great, thanks @emmanuelbernard

PR #985, opened 2023-12-15
 DBZ-7148: Blog post about JDBC Sink connector batch support

mfvitale:


_posts/2023-12-06-JDBC-sink-connector-batch-support.adoc
```diff
@@ -0,0 +1,160 @@
...
+Therefore, the Debezium JDBC connector with batch support is 79 times faster than the version without batch support.
+
+.{nbsp}
+image::1M-different-batch-size-json.png[role=centered-image]
+
+In _Figure 4_, we observe the behavior of the Debezium JDBC connector writing on MySQL with different batch.size.
+Although the difference is notable at the beginning, it becomes apparent that the throughput continues to slow down, and, on average,
+all the `batch.size` configurations take about 7 minutes to process all events.
+
+This raised a concern for us. After conducting a thorough analysis (profiling), we identified that the bottleneck was in event deserialization.
```
vjuranek: Does it really explains the difference between batches at the beginning? If deserialization is the root cause here, I'd expect that there won't be any difference also at the beginning. This looks like there are possibly also other factors which may affect it (my tip would be some kind of buffering somewhere).

If you have any explanation for it, I'd add it here.

vjuranek (commented):
Nice write up - I had one minor comment, just in case you have more details about the differences. One general comment: I'd add units to the axis on all graphs, especially on the time axis.

_posts/2023-12-06-JDBC-sink-connector-batch-support.adoc
mfvitale: Good point. Maybe I think I should better clarify that this strange behavior lead us to investigate and we have found the problem with ser/der that solved the issue of batch.size scalability but not the slow down that can be related to other stuff that we have not investigated.

mfvitale:
> Nice write up - I had one minor comment, just in case you have more details about the differences. One general comment: I'd add units to the axis on all graphs, especially on the time axis.

@vjuranek  Something like that? 
image

_posts/2023-12-06-JDBC-sink-connector-batch-support.adoc
vjuranek: yes, making this more clear would IMHO make the this part more understandable

vjuranek:
> > Nice write up - I had one minor comment, just in case you have more details about the differences. One general comment: I'd add units to the axis on all graphs, especially on the time axis.
> 
> @vjuranek Something like that? 

yes, this would perfect

_posts/2023-12-06-JDBC-sink-connector-batch-support.adoc
mfvitale: @vjuranek  Do you think it's more clear now?

vjuranek (approved):
LGTM now!

mfvitale:
@jpechane I would like to publish it before Xmas, do you think it's better to publish after the release announcement?

jpechane:
@mfvitale IMHO it is independent so you can publish it even today. It is probably better to not publish two posts on the same date.

mfvitale:
> @mfvitale IMHO it is independent so you can publish it even today. It is probably better to not publish two posts on the same date.

Ok, I'll update the date to today and then merge it. Thanks.

PR #999, opened 2024-02-08
Add /operator redirect

Naros:


Naros:
Looks like the redirect shows a "Redirecting" page, which won't likely work with Helm.
Closing for now, we can revisit this using Ruby if we cannot find any other accommodations.

PR #1021, opened 2024-04-15
Update Q2 2024 roadmap

Naros:
@jpechane needs your review.

### DIFF
--- a/roadmap.asciidoc
+++ b/roadmap.asciidoc
@@ -24,44 +24,20 @@ As our engineering capacity is limited, we're focusing our efforts on a single r
 i.e. patch releases for the current stable minor release (e.g. 1.2.1.Final) are done in case of critical bug fixes only.
 ====
 
-== 2.6 - March 2024
-
-* Official MariaDB connector
-* Oracle hybrid mining strategy
-* Oracle XML data type support (column level only)
-* Db2 z/OS and iSeries source connector
-* Unify snapshot modes across all connectors
-* Design user-friendly offset manipulation (i.e, start at a specific position in the transaction logs)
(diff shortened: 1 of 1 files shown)


roadmap.asciidoc
```diff
@@ -24,44 +24,20 @@ As our engineering capacity is limited, we're focusing our efforts on a single r
...
-* Expand Oracle 23c support
-* Oracle OpenLogReplicator support for XML data types
-* Deprecate Oracle 12.x support
-
 == 2.7 - June 2024
 
+* Official MariaDB connector
+* Db2 z/OS source connector
 * Introduce an SPI to reduce memory footprint for multi-tenant databases with identical schemas
 * Read-only incremental snapshots for other relational connectors
```
mfvitale: Should we mention that this will be mainly an R&D activity with eventually some initial implementation in case of positive result from the investigation?

Naros: Which specific item are you referring @mfvitale, the read-only incremental snapshots?  If so, I think that's reasonable.  Would the plan be implementation in 3.0 then or has that not been decided?

mfvitale: > Which specific item are you referring @mfvitale, the read-only incremental snapshots? If so, I think that's reasonable. Would the plan be implementation in 3.0 then or has that not been decided?

Yes I am referring to read-only incremental snapshot. As of now there is no idea on the feasibility so I'll start with some research and see if there is something that we can use for others databases.

Naros: Updated, now lets see what Jiri thinks of what we have :grin:

jpechane:
@Naros LGTM, thanks!

PR #1032, opened 2024-06-03
Update bundles

jpechane:


jpechane:
Relates to https://github.com/debezium/container-images/pull/379

PR #1072, opened 2024-10-03
Test PR Surge Workflow

Naros:


Naros:
So adding `allow_forks: true` was what was needed on the download action for this process to work :heavy_check_mark:

PR #1078, opened 2024-10-25
Update users.asciidoc

gaurav7261:
- Added Zepto company, www.zeptonow.com


Naros:
Thanks @gaurav7261, applied!

PR #1088, opened 2024-11-21
Initial draft of 2025 roadmap

jpechane:


mfvitale:
@jpechane Do we need to further review it or do you think it can be published?

jpechane:
@mfvitale I'd moved forward and published it. We can refine it if needed.
@Naros Do you plan to write a blogpost with some overview?

PR #1094, opened 2024-11-28
New blog post link, "Building a Data Lake with Debezium and Apache Iceberg"

ismailsimsek:
"

### DIFF
--- a/documentation/online-resources.asciidoc
+++ b/documentation/online-resources.asciidoc
@@ -51,6 +51,7 @@ Thanks!
 
 == Blog Posts & Articles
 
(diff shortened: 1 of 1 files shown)


jpechane:
@ismailsimsek Don't you want to re-format that on the one line there will be links to all three parts? So it would look like
Building a Data Lake with Debezium and Apache Iceberg (Part 2, Part 3) by Ismail Simsek

ismailsimsek:
> @ismailsimsek Don't you want to re-format that on the one line there will be links to all three parts? So it would look like Building a Data Lake with Debezium and Apache Iceberg (Part 2, Part 3) by Ismail Simsek

update it

jpechane:
@ismailsimsek Applied, thanks

PR #1126, opened 2025-04-10
Use 3.2 as devel version

vjuranek:


jpechane:
@vjuranek Applied, thanks

PR #1139, opened 2025-06-24
DBZ-9155 new FAQ toc added

ruettere:
ORA-00600: possible solutions added to the documentation FAQ


jpechane:
@ruettere Thanks for the PR. One question - do you think it would be better to place it into Oracle specific https://debezium.io/documentation/reference/3.1/connectors/oracle.html#oracle-frequently-asked-questions ?

ruettere:
> @ruettere Thanks for the PR. One question - do you think it would be better to place it into Oracle specific https://debezium.io/documentation/reference/3.1/connectors/oracle.html#oracle-frequently-asked-questions ?

@jpechane Yes, you're right, it's definitely better placed in the Oracle FAQ. I'm gonna do it ...

PR #1141, opened 2025-07-01
DBZ-9025 Use snapshot script on website builds

Naros:
This is reference to this PR in the main repository
https://github.com/debezium/debezium/pull/6536

### DIFF
--- a/.github/workflows/gh-pages-deployment.yml
+++ b/.github/workflows/gh-pages-deployment.yml
@@ -11,6 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
+      - name: Update snapshot links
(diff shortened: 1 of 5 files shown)


Naros:
So just testing this with the staging workflow, and it works 
https://central.sonatype.com/repository/maven-snapshots/io/debezium/debezium-connector-mariadb/3.2.0-SNAPSHOT/debezium-connector-mariadb-3.2.0-20250701.180408-144-plugin.tar.gz

So if this seems accepted to you @jpechane, then I would suggest we repurpose your PR on the main repository to add a scheduled call to run `website-build-workflow` in the main repo since we have the pre-existing hook there to call into the build here, and that should "just work".

wdyt?

jpechane:
@Naros Yes, I agree. Feel free to make the changes as necessary.

PR #1191, opened 2025-11-20
Add Databricks Debezium blog post to online resources

kavpreetgrewal:
Adds a new Medium blog post about using Debezium to build CDC pipelines on Databricks. Hopefully, this is helpful for users who are looking to do something like this.


mfvitale:
Thanks @KavpreetGrewal!

PR #1202, opened 2025-12-09
Fix postgres-plugin redirect to expected target

ChristopherHolly:
Redirect would result in a 404 because target link uses a hyphen, not underscore, for postgres-plugins.


vjuranek (approved):

vjuranek:
Merged, thanks @ChristopherHolly !

PR #1267, opened 2026-05-28
Fix Postgres 18 support

vjuranek:


Naros (approved):

Naros:
Applied, thanks @vjuranek

PR #1311, opened 2026-09-16
3.7.0.Beta2 announcement

Naros:


mfvitale:
Full of contents!

mfvitale (approved):

Naros:
> Full of contents!

Yep, and it seems our preview releases are getting bigger and bigger. If we keep up this pace, I'm afraid the final release blogs may need a filter pass on new improvements and features to avoid getting far too long.

But that's a good problem to have I would say :wink:

